PR #6, opened 2021-03-17
DNN model training pipeline

dmitryduev:
This PR implements a model training pipeline for deep neural nets. 

The PR contents:
- Training pipeline that is invoked by a new `scope.py` argument: `train`. For example:
```sh
./scope.py train \
  --tag=vnv --path_dataset=data/training/dataset.d15.csv \
  --scale_features=min_max --batch_size=64 --lr=32-4 \
  --epochs=100 --patience=10 \
  --pre_trained_model=models/vnv/20210324_220331/vnv.20210324_220331 \
  --verbose=1 --save
```
All the necessary metadata/configuration could be defined in `config.training`, but could also be overridden with optional arguments, e.g. `./scope.py train ... --batch_size=32 --threshold=0.6 ...`.
  - The pipeline uses the `ScopeNet` models defined in `scope.nn` as subclassed `tf.keras.models.Model` models.
  - `Dataset` class in `scope.utils` that hides the complexity of our dataset handling "under the rug".
- Tools to fetch pre-trained models and datasets from the GCS (requires permissions to access the `gs://ztf-scope` bucket).
- Specify feature name sets in `config.features` to be referenced in `config.training.classes.<class>.features`.
- Specify feature stats in `config.feature_stats` to scale/standardize features before training.
- Refactor the way the taxonomy is described/defined in the config, should be much more verbose. For generating the docs, under the hood, the relevant parts of the specs are recursively converted into something that `tdtax` can render.
- Testing the pipeline end-to-end with a mock dataset.
- Implement GH action to run the test suite.
- Set maximum running time for GH actions to 20 minutes.

The main point of this PR is to make training data/metadata/artifacts easily/consistently "versionable"/trackable so that it only a matter of choosing the tool to do the tracking - that will be implemented in the next PR. (I'm leaning towards [W&B](https://wandb.com) for that part.)


config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+      subclasses:
+        - class: vnv
+          name: variable
+          description: |
+            Any ZTF source that shows detectable variability in its light curve
+            due to an astrophysical origin. This excludes variability due to blended photometry,
+            bright nearby stars, or any CCD artifact.
+        - class: pnp
+          name: periodic
+          description: |
```
AshishMahabal, line 27: Should there be a cut on significance? Or any other criteria? I know that we follow a complex procedure. I was wondering what would be a good place to capture that.

config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+            The ZTF light curve features astrophysical periodic variability.
+            These are typically pulsators, rotating stars and (eclipsing) binaries.
+            This does not include semi-periodic or quasi-periodic variability. This
+            excludes variability due to a varying background (e.g. due to the moon),
+            or spurious periodic variability due to nearby bright stars or other artifacts.
+        - class: longt
+          name: long timescale
+          description: |
+            Any object which shows variability on timescales of 100 days. This can be a steady
```
AshishMahabal, line 36: "100 days or more"? "at least 100 days"? "100+ days"?

config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+            increase/decrease in luminosity, e.g., AGN and CVs.
+            Long timescale period variable like Miras and the more
+            irregular semi-regular variables are generally included in this category.
+        - class: i
+          name: irregular
+          description: |
+            Objects which show irregular variability.
```
AshishMahabal, line 43: irregular/stochastic?

config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+            - class: ea
+              name: detached binary
+            - class: eb
+              name: semi-detached binary
+            - class: ew
+              name: overcontact binary
+        - class: fla
+          name: flaring
+          description: |
+            Any ZTF light curve that shows flares, rapidly rising and fading events.
```
AshishMahabal, line 62: Should M-dwarfs be mentioned by name?

config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+            These are mostly cataclysmic variables, some young stellar objects, some AGN.
+        - class: bogus
+          name: bogus
+          description: |
+            Any light curve that seems variable, but is not due to any astrophysical variability.
```
AshishMahabal, line 67: ... and just bad subtractions?

config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+            These are galaxies that can seem to vary due to PSF variations,
+            image artifacts like ghosts, blended stars, or diffraction spikes.
+    - class: ontological
+      name: ontological
+      description: |
+        What can be said about an object given not just its ZTF light curve
```
AshishMahabal, line 73: " ... given its light curve and additional data"?

config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+      subclasses:
+        - class: puls
+          name: pulsator
+          description: |
+            Any kind of pulsating star.
+          subclasses:
+            - class: dscu
+              name: Delta Scuti
+              description: |
+                Delta Scuti are pulsating A&F main sequence stars. The pulsation period ranges between
```
AshishMahabal, line 83: expand A&F

config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+                and slow decay, to almost sinusoidal light curve shapes.
+            - class: rrlyr
+              name: RR Lyrae
+              description: |
+                RR Lyrae are radial pulsators on the horizontal branch; they are helium core burning and
+                hydrogen shell burning. The pulsation period ranges between 0.2 and 1.0 days.
+            - class: lpv
+              name: long period variable
+              description: |
+                Long Period Variables are cool giant stars. Nearly all stars of this type show
```
AshishMahabal, line 101: 80+ days?

config.defaults.yaml
```diff
@@ -6,41 +6,152 @@ kowalski:
...
+          subclasses:
+            - class: wuma
+              name: W Uma
+              description: |
+                Eclipsing close binary stars of spectral types F, G, or K that share a common envelope
+                of material and are thus in contact with one another.
+            - class: blyr
+              name: Beta Lyr
+              description: |
+                Binary systems were one of the components has evolved into a subgiant or giant star and
```
AshishMahabal, line 126: "where one of"

config.defaults.yaml
```diff
@@ -51,3 +162,137 @@ docs:
       coordinates: [254.7213, 32.8341]
       period: 0.31087445597147706
```
AshishMahabal, line 163: We should avoid excess precision in numbers

config.defaults.yaml
```diff
@@ -51,3 +162,137 @@ docs:
...
       title: RR Lyrae c
+
+features:
```
AshishMahabal, line 166: Since ra/dec are not "features" should they be separated as "metadata"?

config.defaults.yaml
```diff
@@ -51,3 +162,137 @@ docs:
...
+training:
+  dataset: md5 checksum? filename? from GCS? to use for training
+  classes:
+    agn:
+      label: AGN
+      # value should refer to features section of this config
+      features: ontological
+      # our "labels" are floats [0., 0.25, 0.5, 0.75, 1.]
+      threshold: 0.7
+      # balance ration for the prevalent class. leave as null to use all available data
```
AshishMahabal, line 296: "ratio". Where do we store numbers for threshold and balance used in the classifier versions so far?

scope.py
```diff
@@ -152,23 +155,24 @@ def _get_nearest_gaia(
...
         df = pd.DataFrame.from_records(gaia_nearest)
 
         df["M"] = df["phot_g_mean_mag"] + 5 * np.log10(df["parallax"] * 0.001) + 5
-        df["Ml"] = df["phot_g_mean_mag"] + 5 * np.log10((df["parallax"] + df["parallax_error"]) * 0.001) + 5
+        df["Ml"] = (
+            df["phot_g_mean_mag"]
+            + 5 * np.log10((df["parallax"] + df["parallax_error"]) * 0.001)
+            + 5
+        )
         df["BP-RP"] = df["phot_bp_mean_mag"] - df["phot_rp_mean_mag"]
```
AshishMahabal, line 176: Are we using errors on BP/RP in any way? The question that was asked of Laurent in the first ZTF-Gaia talk was important: will they be presenting average colors with error-bars in DR3. He didn't know.

AshishMahabal (commented):
Hi, many of my comments are for Jan as they involve descriptions of the variables. Dima, the first few got logged as "single comments" but the later ones were labeled as "pending". Not sure what that means.

dmitryduev:
@AshishMahabal thanks for the comments! While these are most useful, please don't review while a PR is marked as WIP (work in progress) and/or draft. Once it's ready for review, I remove "WIP" from the PR name, mark it as ready for review, and explicitly request reviews from people. Could be a lot of potentially wasted work, since a lot of the stuff could change.

Please don't post single comments that are meant as a part of review. If you hit "start a review", all your comments will be marked as "pending" until you hit "review changes" -> "submit review" - then they'll all be submitted as a part of your review.

config.defaults.yaml
dmitryduev: Not sure. @mcoughlin what say you, my friend?

config.defaults.yaml
dmitryduev: Fixed.

config.defaults.yaml
dmitryduev: Fixed.

config.defaults.yaml
dmitryduev: Sure, why not. To clarify: I just copy-pasted whatever we have in the paper-I (correcting the naming issues that were previously there).

config.defaults.yaml
dmitryduev: ...which is not astrophysical variability ;)

config.defaults.yaml
dmitryduev: Changed to simply "ZTF data"

config.defaults.yaml
mcoughlin: We have some rules of thumb, but the great thing about pnp is it seems to have solved the significance as a function of period problem. So not something simple we would include (we use ~10 for scanning the folders with Kevin, but that focuses on a single period range).

config.defaults.yaml
dmitryduev: Fixed.

config.defaults.yaml
dmitryduev: Where should we truncate? @mcoughlin?

config.defaults.yaml
dmitryduev: Well, we are using them as features (just as, e.g. info on the cross-matches), so they do belong here.

config.defaults.yaml
dmitryduev: This has been a major issue with how things were set up. One would need to dig trough quite a bit of logs to extract that info. My plan has been: once this PR is merged, incorporate [W&B](wandb.com) support to track all the data/metadata/artifacts. There are other options of course, @stefanv and I have been thinking about that - this is something that we should discuss.

dmitryduev: The main point of this PR is to make things easily "versionable"/trackable so that it only a matter of choosing the tool to do the tracking.

scope.py
dmitryduev: This part of code wasn't touched in this PR, except for linting, so I think this should be moved into issues.
(I think we don't use it in this context).

config.defaults.yaml
dmitryduev: BTW, it's also untouched in this PR, so please file an issue.

dmitryduev:
@AshishMahabal @mcoughlin @doccosmos @janvanroestel ready for review!

scope/nn.py
```diff
@@ -0,0 +1,340 @@
...
+    def call(self, inputs, **kwargs):
+        x = self.conv_0(inputs)
+        for i in range(1, self.repetitions):
+            x = vars(self).get(f"conv_{i}")(x)
+        x = self.max_pool(x)
+
+        return x
+
+
+class ScopeNet(tf.keras.models.Model, ABC):
```
mcoughlin, line 84: This looks at least similar to paper I, so maybe reference that paper for the network discussion?

mcoughlin (approved):

scope/nn.py
dmitryduev: Good call, added. FYI It's the same architecture, I just significantly improved the code quality - no more spaghetti.

scope/utils.py
```diff
@@ -1,14 +1,23 @@
 __all__ = [
+    "Dataset",
+    "forgiving_true",
```
AshishMahabal, line 3: What is this?

AshishMahabal: Ah, I see.

config.defaults.yaml
AshishMahabal: But they are not used as a variable, right? I think we had discussed this in the past. Even if they belong here, should they be designtaed differently?

AshishMahabal (approved):
Looks great.

Some of my older comments for @janvanroestel  are in there, but those are about description/detail, not code.

dmitryduev:
@mcoughlin I'm having some hard time trying to debug my subclassed `tf.keras.models.Model` model, so for now I reverted to simply using Keras'es Functional API, which seems to work fine. I suggest to merge this as is now, and I'll dig deeper into that later. Agreed?

mcoughlin:
Agreed!

PR #23, opened 2021-05-04
Adding CVs to the field guide

cjol245:
Dr. Szkody and Claire's contribution to the field guide on CVs. Includes four CV subtypes: U Gem, Z Cam, SU UMa, and Novalike. Each subtype has a graph; we propose that the size of the graphs be changed from 16,9 to 16,5 (similar to the RR Lyrae and Delta Scuti graphs) in order to better represent the CV characteristics. The change would require altering these lines (54-62) in scope/utils.py:

    fig = plt.figure**(figsize=(16, 9))**
    if title is not None:
        fig.suptitle(title, fontsize=24)

    if period is None:
        ax1 = fig.add_subplot(111)
    else:
        ax1 = fig.add_subplot(211)
        ax2 = fig.add_subplot(212)

If it is too time-consuming or complex, the change is not necessary, but it is preferred.


mcoughlin:
@cjol245 Also good to rebase this to deal with the conflicts when you remove those other files.

dmitryduev (changes requested):
Nice work, thanks @cjol245! A few requests:
- Feel free to implement the necessary changes in `scope/utils.py` yourself (i.e. if period is None, use less space for the plot, otherwise keep what we have now) and post it as a part of this PR.
- Please merge in the latest main `git pull upstream main`, which will likely resolve some of the conflict that your branch has with main.
- Looks like you removed everything from the default config - please add it back together with your changes.
- Remove all png files (`git rm --cached doc/data/*png`) and (`doc/_static/taxonomy.html`), they get generated automatically. In general, I would recommend adding changed files first (`git add <file names>`) and then committing those changes (`git commit -m "<descriptive comment>"`).

cjol245:
@dmitryduev Ok!
Would you recommend that I remove the automatically generated files with git remove, or that I revert the commit using git reset, then add back the necessary files?
What happened with the default config is that I altered and renamed the file config.yaml; I didn't realize that I needed to keep the original. Can I request that this information be added to the Dev Guidelines?

dmitryduev:
`git rm --cached ...` + commit is fine - we anyway squash the commits contained in a PR into a single commit before merging, so that the commit history on `main` does not get clogged up. This also means that you merge in main (and not worry about rebasing) as many times as you want.
Regarding the guidelines, this is what we have now:
> Make sure the structure of your config file is the same as the default, i.e. you propagated the changes in `config.defaults.yaml`. (otherwise the `scope.py` utility will later complain and ask you to fix that).
Feel free to suggest an alternative wording.

`scope.py` only checks if _at least_ what is present in the default config file is in your local `config.yaml`.
Because of the merge conflict, the CI did not run and you could not tell if the tests would pass. Once the conflicts are resolved, the CI would run the automated test suite and you would see the errors.

cjol245:
I see. So scope.py requires config.defaults.yaml to be able to check config.yaml against it.
The documentation is probably fine then. I misunderstood 'propagated' in a coding context, and took it to mean that I should make the changes in config.defaults.yaml rather than in a copy of it. My fault, sorry!

scope/utils-mod.py
```diff
@@ -0,0 +1,154 @@
...
+        1: "#28a745",
+        2: "#dc3545",
+        3: "#00415a",
+        "default": "#f3dc11",
+    }
+
+    mask_good_data = light_curve_data["catflags"] == 0
+    df = light_curve_data.loc[mask_good_data]
+
+    if period is not None:
```
cjol245, line 54: Modifications from lines 54 - 63

scope/utils-mod.py
```diff
@@ -0,0 +1,154 @@
...
+        fig = plt.figure(figsize=(16, 9))
+        ax1 = fig.add_subplot(211)
+        ax2 = fig.add_subplot(212)
+    else:
+        fig = plt.figure(figsize=(16, 5))
+        ax1 = fig.add_subplot(111)
+        
+    if title is not None:
+        fig.suptitle(title, fontsize=24)
+
```
cjol245, line 64: Changes concern lines 54 - 63

dmitryduev:
@cjol245 I fixed the remaining merge conflict so now the CI was able to run. It looks like the pre-commit hook is not set up on your end - linting failed (see below - click on "Details" next to red crosses for the failed actions). Please see https://scope.ztf.dev/developer.html#setting-up-your-environment; you also may want to run `./scope.py lint`.

dmitryduev:
@janvanroestel do you have any comments on the content?

janvanroestel (approved):

dmitryduev (approved):

dmitryduev:
@cjol245 I fixed the merge conflicts for you. It is now green on the CI, so merging since @janvanroestel also approved.

PR #24, opened 2021-05-05
Field guide section on Beta Lyrae

mcoughlin:
So I think things are a bit broken because of gloria having lightcurves and kowalski EDR3... not sure what to do. @dmitryduev ?


config.defaults.yaml
```diff
@@ -55,6 +55,10 @@ docs:
       coordinates: [270.4229, 7.1722]
       period: 0.26901030387060865
       title: W Uma
+    beta_lyr:
+      coordinates: [259.5077971, 22.7978128]
+      period: 0.03979692049952155
```
przemekmroz, line 60: This object has an unusually short period. I suggest including another star with a more typical period (~0.5-a few days).

mcoughlin: @przemekmroz updated!

przemekmroz (approved):

dmitryduev (approved):
LGTM! I suggest to wait a bit until Gaia EDR3 is ingested into Gloria so that all the stuff/plots are properly generated and posted.

PR #25, opened 2021-05-05
Add RA/Dec Plot

mcoughlin:
radec__cepheid

Possible plot for the sky location.


AshishMahabal:
The plot looks great. Can we add some RA/Dec labels?

scope/utils.py
```diff
@@ -151,3 +153,131 @@ def plot_gaia_hr(
...
+        transform=ax.transAxes,
+        weight="bold",
+    )
+    plt.text(
+        0.5, 0.0, r"S", va="top", ha="center", transform=ax.transAxes, weight="bold"
+    )
+
+    color = "k"
+    lw = 10
+    alpha = 1.0
```
dmitryduev, line 266: I'd do 0.75 or something.

dmitryduev (changes requested):
Looks great, just a couple comments:
- You need to configure the pre-commit hook to ignore fits files (see https://github.com/ZwickyTransientFacility/scope/blob/main/.pre-commit-config.yaml).
- Change alpha on the plotted positions, otherwise things can get too crowded.
- I would shade out the part of the sky that is not visible from Palomar, otherwise the plots with our golden sets will be confusing.

mcoughlin:
radec__cepheid

@AshishMahabal @dmitryduev

dmitryduev:
Looks really nice, thanks @mcoughlin! Linting failed for whatever reason, but otherwise should be ready for merging. As said, I just want to wait until EDR3 is ingested into G (should be a couple days - I realized that I am also ingesting AllWISE and PS1 to be able to run x-match when ingesting features).

dmitryduev (approved):

scope/utils.py
```diff
@@ -151,3 +153,143 @@ def plot_gaia_hr(
...
+            pos[1],
+            color=color,
+            marker="o",
+            coord=coordsys,
+            lonlat=True,
+            lw=lw,
+            alpha=alpha,
+            zorder=10,
+        )
+
```
janvanroestel, line 292: we should add an optiion to change the symbol size to avoid overlap of symbols; either if the sample is large or very clustered.

scope/utils.py
```diff
@@ -151,3 +153,143 @@ def plot_gaia_hr(
...
+    if save is not None:
+        fig.tight_layout()
+        plt.savefig(save)
```
janvanroestel, line 295: Looks very good overall

dmitryduev:
Got a verbal 👍🏻 from @janvanroestel, so merging.

PR #29, opened 2021-05-10
Fix issues with the LPV section in Field guide

dmitryduev:


dmitryduev:
This is a quick fix for a small issue deployed into prod, so self-merging.

PR #30, opened 2021-05-12
added line for radec plot

AshishMahabal:
Did this create the RA-Dec plot as needed for the LPVs?


mcoughlin (approved):
Yup, that's all there is too it.

dmitryduev (approved):

PR #33, opened 2021-05-14
Example period histogram plots

mcoughlin:
Period histograms based on the feature set.

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -350,10 +350,7 @@ class Scope:
                     print(f'No features for {golden_set_name}')
                     continue
 
-                if golden_set_name in period_limits:
-                    limits = period_limits[golden_set_name]
-                else:
-                    limits = None
+                limits = period_limits.get(golden_set_name)
 
                 if golden_set_name in period_loglimits:
                     loglimits = period_loglimits[golden_set_name]
--- a/scope.py
+++ b/scope.py
@@ -21,7 +21,7 @@ from scope.utils import (
     plot_gaia_hr,
     plot_light_curve_data,
     plot_gaia_density,
-    plot_periods
+    plot_periods,
 )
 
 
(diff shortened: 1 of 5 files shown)


scope.py
```diff
@@ -278,6 +324,49 @@ def doc(self):
...
+                features = self._get_features(positions=positions)
+
+                if len(features) == 0:
+                    print(f'No features for {golden_set_name}')
+                    continue
+
+                if golden_set_name in period_limits:
+                    limits = period_limits[golden_set_name]
+                else:
+                    limits = None
```
dmitryduev, line 356:
```suggestion
                limits = period_limits.get(golden_set_name)
```

scope.py
```diff
@@ -278,6 +324,49 @@ def doc(self):
...
+        period_limits = {'cepheid': [1.0, 100.0],
+                         'delta_scuti': [0.03, 0.3],
+                         'beta_lyr': [0.3, 25],
+                         'rr_lyr': [0.2, 1.0],
+                         'w_uma': [0.2, 0.8]}
+        period_loglimits = {'cepheid': True,
+                            'delta_scuti': False,
+                            'beta_lyr': True,
+                            'rr_lyr': False,
+                            'w_uma': False}
```
dmitryduev, line 336: Lint, please

scope/utils.py
```diff
@@ -102,6 +102,60 @@ def plot_light_curve_data(
...
+    plt.rc("text", usetex=True)
+
+    # make figure
+    fig, ax = plt.subplots(figsize=(6, 6))
+    if title is not None:
+        fig.suptitle(title, fontsize=24)
+
+    if limits is not None:
+        if loglimits:
+            edges = np.logspace(np.log10(limits[0]), np.log10(limits[1]),
```
dmitryduev, line 126: Lint here and below. How come the linter didn't complain?

dmitryduev (approved):
Nice @mcoughlin. Could you please add a screenshot or two to the description? Otherwise LGTM

scope.py
mcoughlin: What exactly is failing? I ran the pre-commit install and so on.

scope/utils.py
mcoughlin: again, what is failing?

mcoughlin:
period__cepheid

Plots are simple. Can be fancied up as people like down the line.

scope.py
```diff
@@ -350,10 +350,7 @@ def doc(self):
                     print(f'No features for {golden_set_name}')
                     continue
 
-                if golden_set_name in period_limits:
-                    limits = period_limits[golden_set_name]
-                else:
-                    limits = None
+                limits = period_limits.get(golden_set_name)
 
                 if golden_set_name in period_loglimits:
```
dmitryduev, line 355: Also, same as above, please.

scope/utils.py
```diff
@@ -102,6 +102,60 @@ def plot_light_curve_data(
...
+    save: Optional[str] = None,
+):
+    """Plot a histogram of periods for the sample
+
+    """
+    # plot the H-R diagram for 1 M stars within 200 pc from the Sun
+    plt.rc("text", usetex=True)
+
+    # make figure
+    fig, ax = plt.subplots(figsize=(6, 6))
```
dmitryduev, line 120: Make it (16, 5) to be on par with the other plots?

mcoughlin: That's only true of lightcurves. gaia is (6,6) for example. Want me to still do 16, 5?

dmitryduev: Whatever is more pleasing to your experienced eye (mine are not).

scope.py
dmitryduev: I accidentally introduced a bug in the pre-commit hook config in a previous pr. Just fixed, please merge in main, linter will work then.

scope/utils.py
mcoughlin: I think smaller is OK for the histograms here (we don't have much data).

dmitryduev (approved):
LGTM!

przemekmroz (approved):

PR #36, opened 2021-05-17
Add missing sample CV subtypes to config.defaults

dmitryduev:


mcoughlin (approved):
Simple addition of known objects. Only one review required for this. Merging.

PR #53, opened 2021-11-03
Do not use Dropout for input features/dmdt's

dmitryduev:
Randomly dropping input feature/dmdt data is a bad idea and I'm not sure why I had it there in the first place. My guess is that the DNN performance will significantly improve once that is removed.

### DIFF
--- a/.requirements/dev.txt
+++ b/.requirements/dev.txt
@@ -6,6 +6,6 @@ numpy<1.20
 pytest>=6.1.2
 questionary>=1.8.1
 scikit-learn>=0.24.1
-tensorflow>=2.4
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
Embarrassed to say that we should have realized the same on initial review ;)

mcoughlin:
Do you understand the keras complaints @dmitryduev ?

dmitryduev:
I'd guess that tf just lives a very intense life that we should all be envious about. Let's see if my hypothesis is correct.

dmitryduev:
there you go @mcoughlin!

PR #72, opened 2022-04-12
add download classification script

jasonmsetiadi:


mcoughlin:
@jasonmsetiadi why is there more than one script in this PR? Shouldn't your master branch be concurrent with scope master? Can you use a branch instead?

PR #75, opened 2022-04-20
Get ids as a list for a specific quad of a CCD for a particular ZTF f…

AshishMahabal:
…ield.

Can be used later for getting features, and inferencing.

### DIFF
--- a/tools/get_quad_ids.py
+++ b/tools/get_quad_ids.py
@@ -1,0 +1,52 @@
+from penquins import Kowalski
+import json
+
+
+def get_field_ids(catalog, field=301, ccd=2, quad=3, minobs=5, n=10):
+    '''Get ids for a specific quad of a CCD for a particular ZTF field.
+
+    USAGE: data = get_field_ids('ZTF_sources_20210401',field=301,ccd=2,quad=3,minobs=5,n=20)
+    It is assumed that you are connected to the database already.
+    ccd range: [1,16] (not checked)
+    quad range: p1,4] (not checked)
+    Defualt minobs is 5
+    Default number of ids to return is 10 (use 0 for all rows)
+    Output: list of ids
+    '''
+
+    if n == 0:
+        limit = 10000000000
+    else:
+        limit = n
+
+    q = {
+        'query_type': 'find',
+        'query': {
+            'catalog': catalog,
+            'filter': {
+                "field": {"$eq": field},
+                "ccd": {"$eq": ccd},
+                "quad": {"$eq": quad},
+                "nobs": {"$gt": minobs},
+            },
+            "projection": {
+                "_id": 1,
+            },
+        },
+        "kwargs": {"limit": limit},
+    }
+
+    r = gloria.query(q)
+    data = r.get('data')
+    return [data[i]['_id'] for i in range(len(data))]
+
+
+if __name__ == "__main__":
+    # setup connection to gloria to get the lightcurves
+    with open('secrets.json', 'r') as f:
+        secrets = json.load(f)
+    gloria = Kowalski(**secrets['gloria'], verbose=False)
+
+    data = get_field_ids(
+        'ZTF_sources_20210401', field=301, ccd=2, quad=3, minobs=5, n=20
+    )
--- a/tools/get_quad_ids.py
+++ b/tools/get_quad_ids.py
@@ -1,3 +1,4 @@
+import argparse
 from penquins import Kowalski
 import json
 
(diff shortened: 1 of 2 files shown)


tools/get_quad_ids.py
```diff
@@ -0,0 +1,52 @@
+from penquins import Kowalski
+import json
+
+
+def get_field_ids(catalog, field=301, ccd=2, quad=3, minobs=5, n=10):
+    '''Get ids for a specific quad of a CCD for a particular ZTF field.
```
mcoughlin, line 6: Can we have a real numpydocs docstring for this? I.e. of the style
https://github.com/skyportal/skyportal/blob/master/skyportal/handlers/api/observation_plan.py#L481

Should be quick given the info you have.

tools/get_quad_ids.py
```diff
@@ -0,0 +1,52 @@
+from penquins import Kowalski
+import json
+
+
+def get_field_ids(catalog, field=301, ccd=2, quad=3, minobs=5, n=10):
```
mcoughlin, line 5: We also need some kind of skiplimit argument so we can cycle through the field to grab all IDs?

mcoughlin (changes requested):
Looks good @AshishMahabal. I think the main thing missing is a skiplimit type argument so we can loop through and grab like the 100-200 IDs.

tools/get_quad_ids.py
```diff
@@ -0,0 +1,66 @@
...
+    return [data[i]['_id'] for i in range(len(data))]
+
+
+if __name__ == "__main__":
+    # setup connection to gloria to get the lightcurves
+    with open('secrets.json', 'r') as f:
+        secrets = json.load(f)
+    gloria = Kowalski(**secrets['gloria'], verbose=False)
+
+    data = get_field_ids(
```
mcoughlin, line 64: I would suggest we should allow these as command line arguments as we have for the other scripts in tools/

tools/get_quad_ids.py
```diff
@@ -0,0 +1,66 @@
+from penquins import Kowalski
+import json
+
+
+def get_field_ids(catalog, field=301, ccd=2, quad=3, minobs=5, skip=0, limit=10):
```
mcoughlin, line 5: skip = None by default, and then check is None.

mcoughlin (changes requested):
@AshishMahabal Thanks. A few more quick suggestions then should be good to go. @jasonmsetiadi has also been adding brief descriptions of the tool usage in docs/usage, so would be nice to have the same for this one so folks know what capabilities we already have.

tools/get_quad_ids.py
AshishMahabal: Which script in tools? Please point to one, and I will use the same format. Will add details to docs/usage after that.

tools/get_quad_ids.py
mcoughlin: Because you are using it as a None and is much more python for boolean checks.

tools/get_quad_ids.py
mcoughlin: e.g.
https://github.com/ZwickyTransientFacility/scope/blob/main/tools/scope_upload_classification.py#L67

usage file is here:
https://github.com/ZwickyTransientFacility/scope/blob/main/doc/usage.md

tools/get_quad_ids.py
```diff
@@ -0,0 +1,88 @@
...
+    # setup connection to gloria to get the lightcurves
+    with open('secrets.json', 'r') as f:
+        secrets = json.load(f)
+    if args.token:
+        secrets['gloria']['token'] = args.token
+    gloria = Kowalski(**secrets['gloria'], verbose=False)
+
+    #    data = get_field_ids(catalog='ZTF_sources_20210401',limit=args.limit)
+    data = get_field_ids(catalog=args.catalog, limit=args.limit)
+    print(data)
```
mcoughlin, line 88: Maybe better to write it to a file than a print statement?

mcoughlin (approved):
Would be good to write IDs to a file, otherwise looks good.

tools/get_quad_ids.py
AshishMahabal: Ideally this would be part of a flow. For instance in a loop I would call it with all arguments, and get_field_ids "returns" a list that I can capture in whichever way. This shouldn't be used from command line as is anyway. The print by itself is ugly, I agree. I could leave it without assigning to data, and let the user capture it whichever way.

mcoughlin: But if calling from the command line, one should be able to write to a file. I understand if using in a workflow, can call the function and do it however.

PR #76, opened 2022-05-17
Get all quad ids for a ZTF field

saagar-parikh:
Given a CCD and Quad range, get all quad ids in separate hdf5 files (one .h5 for each quad).


tools/get_all_quad_ids.py
```diff
@@ -0,0 +1,256 @@
+import argparse
```
mcoughlin, line 1: Can this just be done as a modification to the original script? i.e. if --all-quads or something, your path is triggered?

tools/output/data.txt
```diff
@@ -0,0 +1,265 @@
+Catalog: ZTF_sources_20210401
```
mcoughlin, line 1: Don't commit the output data files.

tools/secrets.json
```diff
@@ -0,0 +1,8 @@
+{
```
mcoughlin, line 1: I don't think we need this committed either right?

mcoughlin (changes requested):
Not bad:
(i) Don't commit the output files
(ii) Merge your id's script into the existing one and use some command line flag to trigger your code.

.gitignore
```diff
@@ -129,3 +129,4 @@ dmypy.json
 .pyre/
 
 config.yaml
+fritz.yaml
```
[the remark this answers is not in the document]
saagar-parikh: I did that to resolve some pre-commit issue. This file was committed by mistake anyways. I'll remove it.

tools/get_all_quad_ids.py
saagar-parikh: Will do.

PR #79, opened 2022-08-15
Fix bugs in Fritz source uploads

bfhealy:
Modifications and improvements to scope_upload_classification.py, fritz.py, and associated documentation.
-Fixed token authentication for uploading sources
-Fixed bug preventing source assignment to multiple group ids if already in one of the groups
-Revised input syntax for multiple group ids (avoids confusion in Z shell)


mcoughlin (approved):
LGTM!

PR #80, opened 2022-08-15
Improve source uploading and documentation

bfhealy:
Added ability to upload no classification to Fritz, since duplicates easily arise (especially during testing). Also removed sys.path.append() in favor of adding scope directory to PYTHONPATH locally (now described in documentation). Other typo corrections in usage instructions.

### DIFF
--- a/tools/scope_upload_classification.py
+++ b/tools/scope_upload_classification.py
@@ -45,11 +45,15 @@ def upload_classification(
         else:
             # check which groups source is already in
             add_group_ids = group_ids.copy()
-            response = api("GET", f"/api/sources/{obj_id}/groups", args.token)
+            # response = api("GET", f"/api/sources/{obj_id}/groups", args.token)
+            response = api("GET", f"/api/sources/{obj_id}", args.token)
             data = response.json().get("data")
 
+            data_groups = data['groups']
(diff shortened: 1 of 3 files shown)


tools/scope_upload_classification.py
```diff
@@ -62,16 +59,18 @@ def upload_classification(
...
-        json = {
-            "obj_id": obj_id,
-            "classification": classification,
-            "taxonomy_id": taxonomy_id,
-            "probability": prob,
-            "group_ids": group_ids,
-        }
-        response = api("POST", "/api/classification", args.token, json)
+        # allow classification assignment to be skipped (useful to avoid duplicates)
+        if classification != '':
```
mcoughlin, line 63: Don't you need a GET then a POST to check for existing classifications here?

bfhealy: At the moment the new code just lets one upload a source (e.g. to a new group) while knowingly to pass an empty string to avoid duplicate classification. I'll add in a check for existing classifications using GET, since this will be more robust down the line.

mcoughlin: Ok click the review tag thing once you are ready for another look!

mcoughlin (approved):
LGTM

PR #83, opened 2022-08-25
Improvements to scope_upload_classification

bfhealy:
This PR updates scope_upload_classification.py to allow the upload_classification function within to run after import (in addition to being part of an executable script). The PR also adds the ability to skip uploading a period if the value for a given source is missing or NaN. This feature may be useful if the user decides to forego uploading a period associated with a non-periodic label.

### DIFF
--- a/tools/scope_upload_classification.py
+++ b/tools/scope_upload_classification.py
@@ -6,6 +6,7 @@ from penquins import Kowalski
 from time import sleep
 from scope.fritz import save_newsource, api
 import math
+import warnings
 
 
 def upload_classification(
@@ -80,6 +81,9 @@ def upload_classification(
             period = float(row.period)
             if math.isnan(period):
                 period = None
+                warnings.warn('period is NaN - skipping period upload.')
+        else:
+            warnings.warn('period column is missing - skipping period upload.')
 
         # get object id
         response = api("GET", f"/api/sources?&ra={ra}&dec={dec}&radius={2/3600}", token)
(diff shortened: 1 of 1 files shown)


tools/scope_upload_classification.py
```diff
@@ -5,22 +5,36 @@
 from penquins import Kowalski
 from time import sleep
 from scope.fritz import save_newsource, api
+import math
 
 
 def upload_classification(
-    file, gloria, group_ids, taxonomy_id: int, classification: str
+    file: str,
```
mcoughlin, line 12: Good on you for starting to type cast, we should be doing that more.

tools/scope_upload_classification.py
```diff
@@ -58,12 +72,17 @@ def upload_classification(
...
                     probs[cls] = row.iloc[-1 * len(classification) + i]
 
-        ra, dec, period = float(row.ra), float(row.dec), float(row.period)
+        ra, dec = float(row.ra), float(row.dec)
+
+        # Check for and assign period
+        period = None  # default
+        if 'period' in columns:
+            period = float(row.period)
+            if math.isnan(period):
```
mcoughlin, line 81: Maybe throw a warning if a period is NaN?

mcoughlin (approved):
Small comments @bfhealy.

tools/scope_upload_classification.py
bfhealy: Makes sense, I added separate warnings for NaN and missing period column.

PR #85, opened 2022-08-26
Improve classification downloads

bfhealy:
This PR improves the scope_download_classification.py tool. The user can provide a CSV file of object ids and/or ra-dec coordinates to query sources. There is also a "parse" option that queries all sources in a given group(s) without requiring an input file. In each case, the script creates a new CSV file containing each source's classifications, probabilities, and period annotations. The format of the classification column matches the output obtained from a download using the Fritz interface.

### DIFF
--- a/tools/scope_download_classification.py
+++ b/tools/scope_download_classification.py
@@ -6,18 +6,51 @@ from penquins import Kowalski
 from time import sleep
 from scope.fritz import api
 import warnings
+import numpy as np
 
 
 def download_classification(file: str, gloria, group_ids: list, token: str):
     """
     Download labels from Fritz
-    :param file: CSV file containing obj_id column or "all" (str)
+    :param file: CSV file containing obj_id column or "parse" to query by group ids (str)
     :param gloria: Gloria object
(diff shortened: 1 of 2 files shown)


tools/scope_download_classification.py
```diff
@@ -1,82 +1,177 @@
...
-        obj_id = None
-        if data["totalMatches"] > 0:
-            obj_id = data["sources"][0]["id"]
-        print(f"object {index} id:", obj_id)
 
-        # save new source
-        if obj_id is None:
-            obj_id = save_newsource(
-                gloria, group_ids, ra, dec, period=period, return_id=True
+    if (file == "parse") | (file == 'Parse') | (file == 'PARSE'):
```
mcoughlin, line 21: maybe if file in ["parse", 'Parse', 'PARSE']:

mcoughlin (approved):
Very small suggestion. Looks good.

tools/scope_download_classification.py
bfhealy: Agreed, done!

PR #86, opened 2022-08-31
Upload reliability improvements

bfhealy:
This PR improves the stability of source uploads using scope_upload_classification.py:
- All calls to fritz.api() are 'protected' by a try/except loop that catches a JSON error raised by a server response issue. After 10 tries the code will still fail, but this protection helps to keep the loop running in the event of a passing glitch.
- scope_upload_classification now always calls fritz.save_newsource(). New checks run within save_newsource to enable photometry and period data to be posted even if an object already exists on Fritz. This is useful in the case of an interrupted upload, and also if new ZTF photometry becomes available.
- Additional checks prevent the upload of duplicate/nan photometry, which previously raised an error.

### DIFF
--- a/scope/fritz.py
+++ b/scope/fritz.py
@@ -246,6 +246,7 @@ def save_newsource(
     return_id=False,
 ):
 
+    # get the lightcurves
     light_curves = get_lightcurves(gloria, ra, dec, 2.0)
     if len(light_curves) < 1:
         print('No lightcurves found for this objects!')
--- a/scope/fritz.py
+++ b/scope/fritz.py
@@ -274,23 +274,6 @@ def save_newsource(
         )
         obj_id = radec_to_iau_name(ra_mean, dec_mean, prefix="ZTFJ")
 
-        # a source exists on F already?
-        # for attempt in range(10):
-        #    try:
-        #        response = api(
-        #            "GET", f"/api/sources?&ra={ra}&dec={dec}&radius={radius/3600}", token
-        #        )
-        #        data = response.json().get("data")
-        #        break
-        #    except InvalidJSONError:
-        #        print(f'Error - Retrying (attempt {attempt+1}).')
-
-        # if data["totalMatches"] > 0:
-        # print(data)
-        # save source to the groupids if it is already on Fritz
-        #    print("%s already exists on Fritz!" % data["sources"][0]['id'])
-        #    return None
-
         # post new source to Fritz
         if not dryrun:
             post_source_data = {
@@ -321,7 +304,9 @@ def save_newsource(
 
     # get photometry; drop flagged/nan data
     df_photometry = make_photometry(light_curves, drop_flagged=True)
-    df_photometry = df_photometry.dropna().reset_index(drop=True)
+    df_photometry = (
+        df_photometry.dropna().drop_duplicates('expid').reset_index(drop=True)
+    )
 
     # if no photometry exists or there is a difference, post the lightcurves
     if len(data) != len(df_photometry):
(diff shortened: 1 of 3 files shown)


scope/fritz.py
```diff
@@ -293,15 +284,38 @@ def save_newsource(
...
-            response = api("POST", "/api/sources", token, post_source_data)
-            if response.json()["status"] == "error":
-                print(f"Failed to save {obj_id} as a Source")
-                return None
-
-        print("Found %d lightcurves" % len(light_curves))
-
-        # post photometry to obj_id; drop flagged data
-        df_photometry = make_photometry(light_curves, drop_flagged=True)
+            for attempt in range(10):
```
mcoughlin, line 287: Maybe make the number of attempts a variable like 
MAX_ATTEMPTS = 10
Defined at the top of the script?

scope/fritz.py
```diff
@@ -293,15 +284,38 @@ def save_newsource(
...
+                    response = api("POST", "/api/sources", token, post_source_data)
+                    if response.json()["status"] == "error":
+                        print(f"Failed to save {obj_id} as a Source")
+                        return None
+                    break
+                except InvalidJSONError:
+                    print(f'Error - Retrying (attempt {attempt+1}).')
+
+    # start by checking for existing photometry
+    for attempt in range(10):
```
mcoughlin, line 298: Then you can reuse it here.

scope/fritz.py
```diff
@@ -322,10 +336,16 @@ def save_newsource(
 
         if (len(photometry.get("mag", ())) > 0) & (not dryrun):
             print("Attempting to upload as %s" % obj_id)
-            response = api("PUT", "/api/photometry", token, photometry)
-            if response.json()["status"] == "error":
-                print('Failed to post to Fritz')
-                return None
+            for attempt in range(10):
```
mcoughlin, line 339: Here three!

scope/fritz.py
```diff
@@ -334,7 +354,14 @@ def save_newsource(
             "group_ids": group_ids,
             "data": {'period': period},
         }
-        response = api("POST", "api/sources/%s/annotations" % obj_id, token, data=data)
+        for attempt in range(10):
```
mcoughlin, line 357: And four ;)

tools/scope_upload_classification.py
```diff
@@ -35,6 +39,12 @@ def upload_classification(
     sources = pd.read_csv(file)
     columns = sources.columns
 
+    if start is not None:
+        if stop is not None:
+            sources = sources.loc[start:stop]
```
mcoughlin, line 44: Should this be start+1?

tools/scope_upload_classification.py
```diff
@@ -86,45 +96,60 @@ def upload_classification(
             warnings.warn('period column is missing - skipping period upload.')
 
         # get object id
-        response = api("GET", f"/api/sources?&ra={ra}&dec={dec}&radius={2/3600}", token)
-        sleep(0.9)
-        data = response.json().get("data")
+        for attempt in range(10):
```
mcoughlin, line 99: Same variable here.

mcoughlin (changes requested):
I suggest adding a variable at the top of the script to define the number of attempts. Otherwise looks good.

scope/fritz.py
bfhealy: Agreed! I added this variable for both scripts and replaced all range(10) with range(MAX_ATTEMPTS).

tools/scope_upload_classification.py
bfhealy: Since the code currently prints the zero-based index of each row in the dataframe, I wanted to stick with that convention for the start argument. I've noted this convention in usage.md and the help info of the scripts.

bfhealy:
Thanks Michael! I just pushed the changes you requested.

bfhealy:
I also just added handling for a ConnectionError, which also should prompt the MAX_ATTEMPTS loop.

mcoughlin:
LGTM @bfhealy !

bfhealy:
Thanks, and sorry for the last-minute changes - I added handling for a couple more connection-related errors.

mcoughlin:
Which is great. Feel free to merge when ready @bfhealy.

PR #90, opened 2022-09-07
Increase group source download speed

bfhealy:
This PR speeds up group source downloads by eliminating unneeded API calls. If -file is set to 'parse', a download of ~2500 sources completes in ~3 minutes. There remains the option to download specific sources, but this process still loops over each source and takes longer. The new code also adds checkpoints to save results periodically and allows intermediate resumption in the case of an interrupted download.


mcoughlin (approved):
LGTM @bfhealy

PR #95, opened 2022-09-09
Enable multiple classification uploads

bfhealy:
This PR enables multiple classifications to be uploaded together to Fritz. Classifications are uploaded in a batch with size specified by UPLOAD_BATCHSIZE

Other minor changes:
- option to skip photometry search/upload for existing sources to save time
- use of config.yaml to provide Kowalski authentication, instead of requiring secrets.json
- repositioning of the taxonomy map read-in to avoid doing it multiple times
- use of the RADIUS_ARCSEC variable to provide the radius for coordinate searches

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -84,7 +84,7 @@ output: data with new columns appended.
 
 ## Scope Upload Classification
 inputs:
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM @bfhealy

PR #100, opened 2022-09-14
Overhaul taxonomy upload script

bfhealy:
This PR modifies taxonomy.py to facilitate taxonomy uploads to Fritz. Users can now import a taxonomy module (e.g. [scope-phenomenology](https://github.com/bfhealy/scope-phenomenology) in python, pass in the module into taxonomy.py, and upload it. The original functionality requiring a path to a YAML file remains as another option, although users must now also specify a name, version and provenance.

### DIFF
--- a/tools/taxonomy.py
+++ b/tools/taxonomy.py
@@ -26,6 +26,7 @@ def post(
     provenance: Optional[str] = None,
 ):
     """Post taxonomy to Fritz
+       NOTE: token in config.yaml must have 'Post taxonomy' permission
 
     $ ./taxonomy.py \
       --taxonomy='phenomenological.yaml' \
--- a/tools/taxonomy.py
+++ b/tools/taxonomy.py
@@ -84,6 +84,14 @@ def post(
     )
     print(response.json())
 
+    response_json = response.json()
+    if response_json['status'] == 'success':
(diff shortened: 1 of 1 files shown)


tools/taxonomy.py
```diff
@@ -1,45 +1,86 @@
...
-        tax["group_ids"] = list(group_ids)
+        tax_obj["group_ids"] = list(group_ids)
 
     response = fritz_api(
         "POST",
         f"{config['fritz']['protocol']}://{config['fritz']['host']}/api/taxonomy",
-        tax,
+        tax_obj,
     )
     print(response.json())
```
mcoughlin, line 85: Do we need this print statement?

mcoughlin (approved):
Might remove the print statement, otherwise looks good.

tools/taxonomy.py
bfhealy: I think the print statement is helpful in this case because it reports the taxonomy_id associated with the new taxonomy. The id is otherwise somewhat buried in the list of taxonomies returned by a GET request.

mcoughlin: Maybe you should format it and explain what is being printed then?

bfhealy: Sounds good, I'll reformat what's being printed.

PR #101, opened 2022-09-14
Enable classifications from multiple taxonomies

bfhealy:
This PR enables classifications spanning multiple taxonomies to be uploaded to Fritz. golden_dataset_mapper.json, which contains the mapping from dataset labels to Fritz classifications that gets passed into scope_upload_classification.py, is included as a new file in tools. This file also contains the taxonomy ids connected to each classification. Finally, since photometry uploaded to Fritz via the PUT request does not get duplicated, code comparing existing and new photometry in fritz.py has been removed to make uploads run faster.

### DIFF
--- a/scope/fritz.py
+++ b/scope/fritz.py
@@ -303,9 +303,14 @@ def save_newsource(
         df_photometry.dropna().drop_duplicates('expid').reset_index(drop=True)
     )
 
-    # hardcoded this because it is easier, but if Fritz ever changes
-    # this number will change
-    instrument_id = 1
+    # Get up-to-date ZTF instrument id
+    name = 'ZTF'
+    response_instruments = api('GET', 'api/instrument', token)
+    instrument_data = response_instruments.json().get('data')
+
+    for instrument in instrument_data:
+        if instrument['name'] == name:
+            instrument_id = instrument['id']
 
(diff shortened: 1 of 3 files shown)


scope/fritz.py
```diff
@@ -297,58 +297,41 @@ def save_newsource(
...
-            "mjd": df_photometry["mjd"].tolist(),
-            "mag": df_photometry["mag"].tolist(),
-            "magerr": df_photometry["magerr"].tolist(),
-            "limiting_mag": df_photometry["zp"].tolist(),
-            "magsys": df_photometry["magsys"].tolist(),
-            "filter": df_photometry["ztf_filter"].tolist(),
-            "ra": df_photometry["ra"].tolist(),
-            "dec": df_photometry["dec"].tolist(),
-        }
+    # hardcoded this because it is easier, but if Fritz ever changes
```
mcoughlin, line 306: Might be good to do a get and grab the number based on the name?

mcoughlin (approved):
Might check the instrument_id explicitly in a get?

scope/fritz.py
bfhealy: Agreed, I'll implement that change.

PR #102, opened 2022-09-15
Fix batch upload bugs and typos

bfhealy:
This PR fixes bugs relating to batch uploads of classifications:
- The list of classifications is now reset after each upload, preventing duplicate uploads
- The start/stop arguments can now be used independently
- UPLOAD_BATCHSIZE is limited to 10 by default instead of 100 (timeout)
- Batch uploads count up to UPLOAD_BATCHSIZE correctly regardless of start index
- An error is raised if MAX_ATTEMPTS is reached for batch uploads

### DIFF
--- a/scope/fritz.py
+++ b/scope/fritz.py
@@ -327,7 +327,7 @@ def save_newsource(
     }
 
     if (len(photometry.get("mag", ())) > 0) & (not dryrun):
-        print("Uploading photmetry for %s" % obj_id)
+        print("Uploading photometry for %s" % obj_id)
         for attempt in range(MAX_ATTEMPTS):
             try:
                 response = api("PUT", "/api/photometry", token, photometry)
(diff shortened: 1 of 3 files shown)


scope/fritz.py
```diff
@@ -327,7 +327,7 @@ def save_newsource(
     }
 
     if (len(photometry.get("mag", ())) > 0) & (not dryrun):
-        print("Uploading photmetry for %s" % obj_id)
```
mcoughlin, line 330: OOps ;)

tools/golden_dataset_mapper.json
```diff
@@ -36,7 +36,7 @@
 
 "non-variable":
     {"fritz_label": "non-variable",
-      "taxonomy_id:": 1012
+      "taxonomy_id": 1012
```
mcoughlin, line 39: Oh nice catch!

mcoughlin (approved):
Nice.

PR #105, opened 2022-09-15
Always assign start/stop, upload last batch

bfhealy:
The start and stop indices for a dataset are now always assigned in scope_upload_classification.py, even if not specified by the user. Both variables are used later in the script during the batch upload. This PR also fixes a bug in which the final batch of labels does not upload if the batch size is smaller than UPLOAD_BATCHSIZE.


mcoughlin (approved):
LGTM

PR #106, opened 2022-09-16
Additional connection try/except

bfhealy:
This PR 'protects' an api call in fritz.py (added in #101) which previously lacked a try/except loop for connection errors (see #86). The PR also adds new connection-related error (JSONDecodeError) to check in fritz.py and scope_upload_classification.py.


mcoughlin (approved):
LGTM!

PR #108, opened 2022-09-16
Flip sign of magnitude difference

bfhealy:
This PR flips the sign of a magnitude difference to avoid negative `yerr` values.


mcoughlin (approved):
LGTM

PR #109, opened 2022-09-20
Add ZTF id to source downloads

bfhealy:
This PR adds the ability to download fritz annotations containing a ZTF source id using scope_download_classification.py. It effectively adds two columns to the output CSV file containing the annotation origin (e.g. 'ZTF_DR5') and ztf_id.


tools/scope_download_classification.py
```diff
@@ -127,6 +149,8 @@ def download_classification(file: str, gloria, group_ids: list, token: str, star
                 probs += [prb_list]
                 period_origins += [origin_list]
                 periods += [period_list]
+                id_origins += [id_origin_list]
+                ztf_ids += [id_list]
 
             # create dataframe from query results
             sources = pd.DataFrame(
```
guiga004, line 156: Can we explore using the pandas json_normalize function instead of manually populating the dataframe?

https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html

guiga004 (approved):

tools/scope_download_classification.py
guiga004: or https://pandas.pydata.org/docs/reference/api/pandas.read_json.html

bfhealy: @guiga004 Thanks for the suggestion, I'm now using json_normalize and it makes the code more succinct. Let me know what you think of this modification.

guiga004 (approved):

PR #111, opened 2022-09-21
Guiga004 scope download classification robust to server errors - merge fix

guiga004:


scope/fritz.py
```diff
@@ -90,7 +91,19 @@ def api(
         kwargs["json"] = data
     elif method == "GET":
         kwargs["params"] = data
-    response = requests.request(**kwargs)
+
+    for attempt in range(max_attempts):
```
mcoughlin, line 95: There should be some kind of sleep here?

mcoughlin: Probably that sleep should be set as a function argument. Defaulting to 5 seconds?

mcoughlin (approved):
Just need to add some kind of sleep in your retry.

bfhealy (approved):
I agree with Michael's comment, otherwise looks good!

PR #112, opened 2022-09-28
Handle multiple annotation datatypes

bfhealy:
This PR makes scope_manage_annotation.py more robust to different datatypes. If a CSV file is passed in, the annotation column's datatype will be converted from a numpy to python dtype to avoid an error on posting. If a single source and key/value pair is instead specified, the value (initially a str) will be converted to a float or int as appropriate. The script also now loops over all keys under a given origin to check for matches, rather than selecting a single key.


mcoughlin (approved):
LGTM

PR #113, opened 2022-09-29
Revise uploads and increase speed

bfhealy:
This PR revises source uploads by:
- Removing unneeded try/except loops (thanks to #111)
- Adding an optional upload of each source's ZTF id as an annotation
- Selecting the most recently posted source whose id begins with 'ZTFJ' as the existing source (if applicable). Previously, if multiple nearby sources existed, the script would effectively select one at random.
- Updating the usage documentation

fritz.py also has a one-line change to api(): the call to sleep() now takes place after an exception is caught. This change speeds up api calls that do not produce an error.


mcoughlin (approved):
LGTM

PR #114, opened 2022-10-03
Fix loop bug in get_field_features

bfhealy:
This PR fixes a bug that threw an error when using the `get_field_features` function within get_features.py. The criterion to break the loop has been modified so it doesn't fail when the length of the input list of source ids is an integer multiple of the per-query limit. Both get_features functions also now return their main outputs.

### DIFF
--- a/tools/get_features.py
+++ b/tools/get_features.py
@@ -107,6 +107,8 @@ def get_features(
         print(df)
         print("dmdt shape", dmdt.shape)
 
+    return df, dmdt
+
 
 # get features of all ids in a field
 def get_field_features(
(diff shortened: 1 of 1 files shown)


tools/get_features.py
```diff
@@ -107,6 +107,8 @@ def get_features(
         print(df)
```
mcoughlin, line 107: Should this print statement maybe have more context?

tools/get_features.py
```diff
@@ -134,14 +136,12 @@ def get_field_features(
     df_collection = []
     dmdt_collection = []
     while 1:
```
mcoughlin, line 138: Maybe while True by convention?

mcoughlin (approved):
Just a few minor things.

bfhealy:
Thanks Michael, `while 1` is now `while True` and the `print(df)` statement is now also prints "Features dataframe: "

Issue #115, opened 2022-10-03
Combine get_features and get_field_features


bfhealy:
It appears that the main difference between these two functions in get_features.py is the `outfile` variable containing the path to the directory for saved results. I think these functions can be combined into one that uses the existing `whole_field` keyword to assign the correct path.

Issue #58, opened 2022-01-21
script to download labels from Fritz and generate a trainingset


janvanroestel:
The goal is to have a python script to downloads all available labels from Fritz and collects their features and puts it in an easy to use .csv file. The format of this file should be the same as the older d15 trainingset.

mcoughlin:
We need to assess to what extent https://github.com/ZwickyTransientFacility/scope/blob/main/tools/scope_download_classification.py
is sufficient for this. It does not download the features from kowalski, which can be queried as here:
https://github.com/mcoughlin/ztfperiodic/blob/master/ztfperiodic/utils.py#L1789

This should be made to be an option in the code.

bfhealy:
I think it's a good idea to incorporate Saagar's incoming get_features.py code (#84) into scope_download_classification.py to achieve this goal.

PR #118, opened 2022-10-06
Consolidate get_features functions

bfhealy:
This PR consolidates two functions into one inside of get_features.py, resolving #115. It also introduces the write_results keyword to make writing the results optional. This will be useful in addressing #58, where we want to write a combined features/classifications file.


mcoughlin (approved):
LGTM

PR #119, opened 2022-10-07
Enable source/feature merging

### LINKED ISSUE #58: script to download labels from Fritz and generate a trainingset

bfhealy:
This PR incorporates the get_features.py code into scope_download_classification.py to query and download features from Kowalski before merging them with a source list from Fritz, either downloaded or provided as input (closes #58). The merge is done by matching ZTF id numbers after checking that both tables come from the same data release. 

All files produced by scope_download_classification are now saved in a directory called 'fritzDownload' within scope/tools. Previously they were saved in the tools directory.

My tests of the the code produce the same form of file as the initial golden dataset, with minor differences in columns. The newly produced dataset has an additional column called 'coordinates' from the features database, and it is missing the 'nice' and 'niice' classification columns that we have not uploaded to Fritz. The 'zvm_id' column is replaced with 'obj_id' containing the coordinate-based names of each object from Fritz.

### DIFF
--- a/tools/scope_download_classification.py
+++ b/tools/scope_download_classification.py
@@ -222,7 +222,9 @@ def download_classification(
     outpath = os.path.join(os.path.dirname(__file__), 'fritzDownload')
     os.makedirs(outpath, exist_ok=True)
 
-    filename = file.removesuffix('.csv') + '_fritzDownload' + '.csv'  # rename file
+    filename = (
+        os.path.basename(file).removesuffix('.csv') + '_fritzDownload' + '.csv'
+    )  # rename file
     filepath = os.path.join(outpath, filename)
 
     if file in ["parse", 'Parse', 'PARSE']:
@@ -249,6 +251,7 @@ def download_classification(
         else:
             start = 1
             print(f'Downloading {allMatches} sources...')
+
         # iterate over all pages in results
         for pageNum in range(start, pages + 1):
             print(f'Page {pageNum} of {pages}...')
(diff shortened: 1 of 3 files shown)


tools/scope_download_classification.py
```diff
@@ -80,17 +80,152 @@ def organize_source_data(src: pd.DataFrame):
     return dct
 
 
-def download_classification(file: str, group_ids: list, start: int):
+def merge_sources_features(sources, features_catalog, features_limit=1000):
+
+    outpath = os.path.join(os.path.dirname(__file__), 'fritzDownload')
```
mcoughlin, line 85: Maybe this should be made as an input argument? Defaulting to this?

tools/scope_download_classification.py
```diff
@@ -80,17 +80,152 @@ def organize_source_data(src: pd.DataFrame):
...
+    if dup_ztf_id > 0:
+        print(f'Dropping {dup_ztf_id} sources with duplicate ZTF ids.')
+        sources = sources.drop_duplicates('ztf_id').reset_index(drop=True)
+
+    features_ztf_dr = features_catalog.split('_')[-1]
+    source_ids = sources['ztf_id'].values.astype(int).tolist()
+
+    # Open golden dataset mapper
+    mapper_dir = os.path.dirname(__file__)
+    mapper_path = os.path.join(mapper_dir, 'golden_dataset_mapper.json')
```
mcoughlin, line 107: Is this also best hard coded like this?

tools/scope_download_classification.py
```diff
@@ -80,17 +80,152 @@ def organize_source_data(src: pd.DataFrame):
...
+    df = df.drop('_id', axis=1)
+
+    # Merge on ZTF id
+    merged_set = pd.merge(expanded_sources, df, on='ztf_id')
+
+    # Make ztf_id last column in dataframe
+    ztf_id_col = merged_set.pop('ztf_id')
+    merged_set['ztf_id'] = ztf_id_col
+
+    filepath = os.path.join(outpath, 'merged_classifications_features.csv')
```
mcoughlin, line 197: Same comment here

mcoughlin (changes requested):
It would be nice to clarify the situation with (or at least document) the paths to the various files.

bfhealy:
I agree about giving the option to change the directory/file names. I've added three new input arguments to allow this to be done.

mcoughlin (approved):
LGTM. Thanks @bfhealy

PR #121, opened 2022-10-10
Handle unrecognized classifications when merging features

bfhealy:
This PR modifies the feature-merging part of scope_download_classification.py to handle classifications that are not in the input dataset label mapper. This issue arises, for example, when a source already existed on Fritz with classifications from a separate taxonomy prior to the user's upload of that source. The new code skips over classifications that are missing from the mapper, which the user can modify if the additional classifications are of interest.


mcoughlin (approved):
LGTM

PR #122, opened 2022-10-11
Consolidate get_quad_ids functions

bfhealy:
This PR consolidates two wrapper functions into one within get_quad_ids.py. It also modifies the usage of the script with the following changes:
- The `--whole-field` and `-all-quads` (now `-multi-quads`) keywords are no longer constantly `True`. They both default to `False` if not included in the script call.
- The `--ccd-range` and `--quad-range` keywords take two integers to allow ranges not starting at 1.
- Kowalski token authentication no longer requires secrets.json, using config.yaml instead
- More status updates printed during run
- `#!/usr/bin/env python` allows running the script with `./`

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -1,14 +1,30 @@
 # Usage
 
-## Download ids for a ZTF quad
+## Download ids for ZTF fields/CCDs/quadrants
 
-```python
-get_field_ids.py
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
@bfhealy Good refactor!

PR #124, opened 2022-10-12
Debug inference code

bfhealy:
This PR makes minor changes, mostly to inference.py:
- The local make_model() function is replaced with a call to nn.DNN.build_model()
  - make_model remains as commented code in case we need to use it for more testing with existing trained models (it contains the initial dropout that has since been removed from the DNN)
- All saved files are put in the same `preds` directory
- default whole_field keyword is now False to be consistent with other parts of our code
- docstring contains accurate kwarg names

### DIFF
--- a/tools/inference.py
+++ b/tools/inference.py
@@ -88,7 +88,8 @@ def clean_data(
     # file to store flagged ids and features with missing values
     if not whole_field:
         filename = (
-            "../preds/field_"
+            BASE_DIR
+            + "/../preds/field_"
             + str(field)
             + "/ccd_"
             + str(ccd).zfill(2)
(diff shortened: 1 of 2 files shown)


tools/inference.py
```diff
@@ -85,10 +88,16 @@ def clean_data(
     # file to store flagged ids and features with missing values
     if not whole_field:
         filename = (
-            "preds/ccd_" + str(ccd).zfill(2) + "_quad_" + str(quad) + "/flagged.json"
```
mcoughlin, line 88: Should we start doing absolute paths?

tools/inference.py
```diff
@@ -85,10 +88,16 @@ def clean_data(
...
+            + str(field)
+            + "/ccd_"
+            + str(ccd).zfill(2)
+            + "_quad_"
+            + str(quad)
+            + "_flagged.json"
         )
     else:
-        filename = "preds/field_" + str(field) + "/flagged.json"
+        filename = "../preds/field_" + str(field) + "/flagged.json"
```
mcoughlin, line 100: Same here.

mcoughlin (approved):
Just a question on absolute paths.

bfhealy:
Agreed, the code now uses BASE_DIR at the beginning of each path.

mcoughlin:
@bfhealy Good to go!

PR #126, opened 2022-10-13
Add all phenomenological/ontological classes to config.defaults.yaml training section

bfhealy:
This PR updates the training section of config.defaults.yaml to include all phenomenological and ontological classes that are present in the golden dataset. While some classes do not currently have enough examples to effectively train the DNN, it is helpful to have a complete list in the config file.


mcoughlin (approved):
LGTM

Issue #130, opened 2022-10-17
Add header to training sets


bfhealy:
To ensure consistency, especially among multiple users training models, we should add a header to each downloaded training set that specifies the scope code version and time of download from Fritz. Other suggestions for the header are also welcome.

PR #132, opened 2022-10-18
Add versioning

bfhealy:
This PR adds versioning to `scope/__init__.py` based on code adapted from [this file](https://github.com/skyportal/skyportal/blob/main/skyportal/__init__.py) in SkyPortal. I think a logical initial version is 0.2.dev0, recognizing development since the initiation of the project. The current git commit hash is also included for further version detail.

The broader goal is to include the code version in a header attached to downloaded training sets (#130).

### DIFF
--- a/scope/__init__.py
+++ b/scope/__init__.py
@@ -1,2 +1,36 @@
 from .nn import *
 from .utils import *
+
+__version__ = '0.2.dev0'
+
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Great idea @bfhealy.

PR #140, opened 2022-10-25
Add Parquet, HDF5 support to Dataset class

bfhealy:
This PR enables .parquet and .h5 files to be read by the `Dataset` class, which is used during training.

### DIFF
--- a/scope/utils.py
+++ b/scope/utils.py
@@ -510,8 +510,8 @@ class Dataset(object):
         verbose: bool = False,
         **kwargs,
     ):
-        """Load csv file with the dataset containing both data and labels
-        As of 20210317, it is produced by labels*.ipynb - this will likely change in a future PR
+        """Load parquet, hdf5 or csv file with the dataset containing both data and labels
+        As of 20221025, it is produced by scope_download_classification.py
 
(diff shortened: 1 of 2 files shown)


scope/utils.py
```diff
@@ -505,30 +505,46 @@ class Dataset(object):
...
-        path_dataset: str,
+        path_dataset: Union[str, pathlib.Path],
         features: tuple,
         verbose: bool = False,
         **kwargs,
     ):
-        """Load csv file with the dataset containing both data and labels
-        As of 20210317, it is produced by labels*.ipynb - this will likely change in a future PR
+        """Load parquet, hdf5 or csv file with the dataset containing both data and labels
+        As of 20221025, it is produced by scope_download_classification.py
```
mcoughlin, line 514: Is this comment important?

mcoughlin (approved):
Only question is on whether that date comment is useful.

scope/utils.py
bfhealy: Probably not as important now since our usage documentation describes how to download and merge labels and features. I'll go ahead and remove it.

PR #142, opened 2022-10-31
Remove ra/dec/field/ccd/quad from training

bfhealy:
This PR updates config.defaults.yaml to exclude a source's ra, dec, and field/ccd/quadrant numbers from training. The phenomenological and ontological classes now share the features in the first list below, and the ontological classes are also trained on the features in the second list.

Screen Shot 2022-10-31 at 10 35 07 AM
Screen Shot 2022-10-31 at 10 35 20 AM


AshishMahabal:
Of the suggested ontological features, we had decided that the following should not be included as "features" (6 Gaia, 1 PS1):

1. Gaia_EDR3__astrometric_excess_noise
2. Gaia_EDR3__parallax_error
3. Gaia_EDR3__pmra
4. Gaia_EDR3__pmra_error
5. Gaia_EDR3__pmdec
6. Gaia_EDR3__pmdec_error
7. PS1_DR1__qualityFlag

In addition, I am of the opinion that the following five belong better to ontological since their derivation is a bit more subjective:

1. period
2. significance
3. pdot
4. n_ztf_alerts
5. mean_ztf_alert_braai

bfhealy:
@mcoughlin I've made more changes following Ashish's recommendations, with the exception that since `pdot` is zero for all training set sources, I continue to exclude it from the feature lists. What do you think of these changes?

mcoughlin:
@bfhealy i think we just try it and see how things look!

bfhealy:
@mcoughlin Ok, sounds good!

bfhealy:
This PR is superseded by #164; the changes discussed here have been implemented.

PR #143, opened 2022-10-31
Add training script

bfhealy:
This PR adds an example training script for all classes. Classes with fewer than 100 positive examples in the golden dataset are commented in the script.


mcoughlin (approved):
LGTM. Is this generated by hand or do we have a script to make it?

bfhealy:
Right now it's made by hand. Might be a good idea to generate it going forward though.

PR #146, opened 2022-10-31
Add method to generate training script

### LINKED ISSUE #144: Code to generate training script
It would be helpful to have Python code to generate a shell script for training (e.g. [train_dnn.sh](https://github.com/ZwickyTransientFacility/scope/blob/main/train_dnn.sh)). This code should take as input the training set, config file, and a keyword argument specifying the minimum number of positive examples needed to perform training. As output, it should generate the training script containing each class with the required number of examples.

bfhealy:
This PR modifies scope.py, adding a method to generate a training shell script in the format of train_dnn.sh. Closes #144

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -17,7 +17,7 @@ from tdtax import taxonomy  # noqa: F401
 from typing import Optional, Sequence, Union
 import yaml
 
-from scope.utils import forgiving_true, load_config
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM!

PR #148, opened 2022-11-01
Add group_ids to photometry json

bfhealy:
This PR adds the `group_ids` keyword to the photometry JSON object that gets uploaded to Fritz. This change will allow photometry to be seen by members of the groups specified by this keyword.


mcoughlin (approved):
LGTM!

PR #149, opened 2022-11-02
Add custom keywords to training script

bfhealy:
This PR allows custom keywords to be appended in bulk to the training script generated using `./scope.py create_training_script`. The usage documentation is also updated to mention this method.


mcoughlin (approved):
LGTM. I would suggest in the docstring also giving examples, as add_keywords is hard to understand what kind of things you would put in it.

bfhealy:
Agreed, I added two examples to the docstring.

PR #152, opened 2022-11-04
Save/load trained models consistently

bfhealy:
This PR simplifies the saving and loading of trained DNN models, which now uniformly use HDF5 format. There was previously an inconsistency between the use of tensorflow format for saving weights (without layers) after training and HDF5 format for loading weights and/or layers before inference. Although saving the weights and layers together takes up more disk space, this approach avoids potential confusion if layers of the model change over time.


mcoughlin (approved):
LGTM

PR #153, opened 2022-11-04
Fix bug when saving individual ccd/quad combos

bfhealy:
This PR fixes a bug which saved the cumulative list of source ids instead of the individual constituents of a given ccd/quadrant when setting `--whole-field = False`. In other words, each subsequent file contained the ids of all previously saved files.

This PR also moves the file-saving process when `--whole-field = False` from the wrapper to the get_field_ids function, making it easier to save one file containing a single ccd/quad combination.


mcoughlin (approved):
LGTM!

PR #154, opened 2022-11-07
Make CSV optional in get_features

bfhealy:
This PR adds a keyword controlling the writing of features as a CSV file, making it optional (and `False` by default) to save disk space. Features continue to be written in .parquet format by default.


mcoughlin (approved):
LGTM

PR #157, opened 2022-11-08
Checkpoint feature downloads

bfhealy:
This PR refactors get_features.py to enable checkpoint saving for large feature downloads. For fields with millions of sources, the feature downloads often freeze and fail. This new code (mainly the `get_features_loop` wrapper) writes features to a parquet file (CSV optional) for every 100,000 sources by default. If the download is cut off, the user can re-run the same code to continue downloading from the last checkpoint.

The existing use of get_features.py code in scope_download_classification.py is tested and adjusted to continue working properly.

### DIFF
--- a/tools/get_features.py
+++ b/tools/get_features.py
@@ -41,7 +41,7 @@ def get_features_loop(
     ccd: int = 1,
     quad: int = 1,
     limit_per_query: int = 1000,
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #158, opened 2022-11-08
Add readme instruction for generating docs locally

bfhealy:
This PR updates README.md to include how to generate local HTML documentation since scope.ztf.dev is down.


mcoughlin (approved):
Thanks @bfhealy

PR #159, opened 2022-11-09
Save multiple feature files per field

bfhealy:
This PR modifies get_features.py to save multiple feature files per field, quantified by a default number of sources (100,000). It removes the prior functionality that updated existing files - reading and appending to increasingly larger files for populous fields becomes very memory-intensive.

The Parquet format supports datasets spread across multiple files (see e.g. https://arrow.apache.org/docs/python/dataset.html#dataset). Going forward, we will be able to perform inference on these file collections using a simple loop to preserve memory.

### DIFF
--- a/tools/get_features.py
+++ b/tools/get_features.py
@@ -97,31 +97,10 @@ def get_features_loop(
             limit_per_query=limit_per_query,
         )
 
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #162, opened 2022-11-10
Loop over multiple parquet feature files during inference

bfhealy:
This PR modifies inference.py to loop over the contents of multiple parquet files containing a field's features. The new code requires much less memory without a major increase in inference time.


mcoughlin (approved):
LGTM

PR #164, opened 2022-11-11
Restructure features lists in config defaults

bfhealy:
This PR continues the debug/improvement process for the inference pipeline. It modifies config.defaults.yaml, restructuring the phenomenological and ontological feature lists. Each feature on the list now has a boolean `include` field (taking the place of the `#` used in the previous version). This PR would supersede #142, so features having `include = True` are the same as agreed upon in that discussion.

Each feature in the config file also has a `dtype` field to explicitly specify the datatype. An upcoming PR will make use of this field to ensure more consistency in files generated by `get_features.py`. Currently, pandas does not always infer types uniformly.

Lines in `scope.py` and `inference.py` where the config features list is used are refactored in this PR for compatibility with the new config format.

### DIFF
--- a/config.defaults.yaml
+++ b/config.defaults.yaml
@@ -771,6 +771,16 @@ features:
     PS1_DR1__zMeanPSFMagErr:
       include: true
       dtype: float
+    _id:
+      include: false
+      dtype: int
+    coordinates:
+      include: false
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #165, opened 2022-11-12
Set feature datatypes and resume interrupted downloads

bfhealy:
Following up on #164, this PR modifies get_features.py to set feature datatypes based on the contents of the scope config file, ensuring consistency among each parquet file saved for the field.

This PR also updates how interrupted downloads are resumed given the use of many parquet files to comprise a field's dataset. As long as DEFAULT_SAVE_BATCHSIZE is kept constant for the field, an interrupted download can be resumed by running the same command again.


mcoughlin (approved):
LGTM

PR #166, opened 2022-11-14
Save inference results in HDF5 format

### LINKED ISSUE #160: Save inference results in HDF5 format
We currently save the results from inferencing as CSV files. It would be more beneficial to write them to HDF5 format, including relevant metadata about the inference (e.g. date, name of trained model file)

bfhealy:
This PR saves inference results in HDF5 format including relevant metadata. CSV files can optionally be saved as well.

Also, the datatype of the `_id` field is changed from pandas `Int64` to `int` to avoid an error when saving inference results. Finally, the version of the code is bumped.

Closes #160

### DIFF
--- a/config.defaults.yaml
+++ b/config.defaults.yaml
@@ -773,7 +773,7 @@ features:
       dtype: float
     _id:
       include: false
-      dtype: Int64
+      dtype: int
     coordinates:
       include: false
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #170, opened 2022-11-15
Fix inference bugs

bfhealy:
This PR fixes outstanding bugs in the inference pipeline:
- Sources with missing features are logged in `missing_dict` with ids in `str` format
- Features in `float64` format are converted to `float32` to avoid a tensorflow error
- Class predictions are merged with an existing file if applicable

### DIFF
--- a/tools/inference.py
+++ b/tools/inference.py
@@ -215,6 +215,10 @@ def run(
         whether to print progress
     time: bool
         print time taken by each step
+    write_csv: bool
+        if True, write CSV file in addition to HDF5
+    float_convert_types: 2-tuple of ints (16, 32 or 64)
+        Existing and final float types for feature conversion
     Returns
     =======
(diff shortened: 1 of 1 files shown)


tools/inference.py
```diff
@@ -388,6 +388,10 @@ def run(
             )  # 1
 
             ts = time.time()
+            # Convert float64 to float32 to satisfy tensorflow requirements
+            features[
+                features.select_dtypes(np.float64).columns
+            ] = features.select_dtypes(np.float64).astype(np.float32)
```
mcoughlin, line 394: Should 64 vs 32 be configurable?

mcoughlin (approved):
just one question on the typing.

tools/inference.py
bfhealy: That makes sense, just pushed a new commit with the change.

PR #171, opened 2022-11-15
Include ra/dec/period columns in inference result

bfhealy:
This PR anticipates uploading inference results to Fritz by including ra, dec and period columns in the files produced by inference.py. The resulting .h5 and .csv files will have all the necessary data for the upload process.


mcoughlin (approved):
LGTM

PR #174, opened 2022-11-18
Add probability threshold for classification uploads

bfhealy:
This PR adds a probability threshold argument to scope_upload_classification.py. Only classifications with a probability greater than the threshold will be uploaded. This is useful for active learning purposes and any time we want to post only a subset of potential classifications.

### DIFF
--- a/tools/scope_upload_classification.py
+++ b/tools/scope_upload_classification.py
@@ -84,9 +84,13 @@ def upload_classification(
 
         if read_classes:
             row_classes = row[classes]  # limit current row to specific columns
-            threshold_keys = row_classes.keys()[
-                row_classes >= p_threshold
-            ]  # determine which dataset classifications are nonzero
+            if p_threshold > 0.0:
+                threshold_keys = row_classes.keys()[
+                    row_classes >= p_threshold
+                ]  # determine which dataset classifications are nonzero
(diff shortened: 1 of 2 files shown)


tools/scope_upload_classification.py
```diff
@@ -82,11 +84,11 @@ def upload_classification(
 
         if read_classes:
             row_classes = row[classes]  # limit current row to specific columns
-            nonzero_keys = row_classes.keys()[
-                row_classes > 0
+            threshold_keys = row_classes.keys()[
+                row_classes > p_threshold
```
mcoughlin, line 88: >= ?

tools/scope_upload_classification.py
```diff
@@ -278,6 +280,13 @@ def upload_classification(
 
     parser.add_argument("-ztf_origin", type=str, help="Origin of uploaded ZTF data")
 
+    parser.add_argument(
+        "-p_threshold",
+        type=float,
+        default=0.0,
+        help="Classification probability > this number to upload",
```
mcoughlin, line 287: >= ?

mcoughlin (approved):
I think better is >= rather than >. Otherwise, LGTM!

bfhealy:
@mcoughlin That sounds good - I just made the changes!

PR #175, opened 2022-11-18
Convert ZTF ID to int before uploading

bfhealy:
This PR ensures that a source's ZTF ID is converted to an integer before its string representation is uploaded to Fritz.


mcoughlin (approved):
LGTM

PR #177, opened 2022-11-22
Fix annotation bug for existing sources with no ZTF ID

bfhealy:
This PR fixes a bug in scope_upload_classification.py where an error is raised during annotation posting if a source exists on Fritz but without any ZTF ID annotation. The code now handles this scenario by posting a new annotation to Fritz in the same way that it would if the source did not exist.


mcoughlin (approved):
LGTM

PR #78, opened 2022-08-09
Hard-code fritz.science URL

bfhealy:
Make new user experience easier by hard-coding https protocol and Fritz URL.

PR #178, opened 2022-11-22
Reorganize and update defaults for API try/except loop

bfhealy:
This PR reorganizes and updates the default values for the try/except loop governing Fritz API calls. The defaults of 10 attempts and 5 seconds between attempts are defined at the top of fritz.py, and one extraneous default value is removed from scope_upload_classification.py

### DIFF
--- a/tools/scope_upload_classification.py
+++ b/tools/scope_upload_classification.py
@@ -11,7 +11,6 @@ import pathlib
 from tools import scope_manage_annotation
 from datetime import datetime
 
-MAX_ATTEMPTS = 10
 RADIUS_ARCSEC = 2
 UPLOAD_BATCHSIZE = 10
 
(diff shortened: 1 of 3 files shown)


tools/scope_upload_classification.py
```diff
@@ -11,7 +11,6 @@
 from tools import scope_manage_annotation
 from datetime import datetime
 
-MAX_ATTEMPTS = 10
```
mcoughlin, line 14: Should this be imported as a default from the config or something?

bfhealy: I think that makes more sense - I'll find a spot in the config to add these defaults.

bfhealy: @mcoughlin Just pushed those changes. I'd also like to undo #78 by reading the protocol and host from the config file again. The multiple times where I've used Scope code to interact with my localhost instance of SkyPortal have highlighted the convenience of being able to switch between protocols/hosts just by editing the config file.

mcoughlin:
@bfhealy Sounds good. Feel free to merge.

PR #182, opened 2022-11-28
Add DNN active learning mapper

bfhealy:
Uploading DNN classifications to Fritz for active learning requires a JSON file to map the column names of inference results to Fritz taxonomy labels. This PR adds the current mapper for DNN active learning to the scope repository.

PR #183, opened 2022-11-28
Standardize taxonomy map keywords, add documentation

bfhealy:
This PR revises `scope_download_classification.py`, renaming the `mapper_name` argument to `taxonomy_map`. This revision achieves consistency with the identical argument in `scope_upload_classification.py`. The documentation in `usage.md` is also updated with a new section describing how to create taxonomy map files. This PR is separate from #182 because it involves more than just a documentation update.


mcoughlin (approved):
LGTM!

PR #187, opened 2022-11-30
Replace model.predict in inference loop to avoid memory leak

bfhealy:
This PR fixes an issue where `model.predict` freezes mid-inference. The issue likely arises due to a memory leak (e.g. [here](https://github.com/keras-team/keras/issues/13118) and [here](https://github.com/tensorflow/tensorflow/issues/44711)) as our code loops over feature files. Among many proposed solutions, the one I've found to make the most difference is replacing `model.predict(features)` with `model(features, training=False)`. This replicates the output of the original code with negligible difference in speed, but prevents ballooning memory to the point of freezing.

### DIFF
--- a/tools/inference.py
+++ b/tools/inference.py
@@ -414,6 +414,13 @@ def run(
                 float_type_dict[float_final]
             )
             # preds = model.predict([features[feature_names].values, dmdt])
+            # Above: calling model.predict(features) in a loop leads to significant
+            #        memory leak and eventual freezing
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM. I would suggest some comment in the code explaining why this helps (not obvious why this fixes it!).

bfhealy:
@mcoughlin Sounds good, I'll add a comment before merging!

PR #188, opened 2022-12-01
Add method to select sources/classifications for active learning

bfhealy:
This PR adds a `select_al_sample` method to `scope.py` to select sources and their classifications from a collection of predictions and make an active learning sample. The method optionally removes sources in the current training set before selecting at random a user-specified minimum number of sources having classifications with probabilities ≥ a given probability threshold. The output is a dataset that can be posted to Fritz using `scope_upload_classification.py`. 

One additional feature to consider is a fixed random state to allow reproducible active learning samples to be made from the same collection of predictions.

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -16,7 +16,8 @@ import tdtax
 from tdtax import taxonomy  # noqa: F401
 from typing import Optional, Sequence, Union
 import yaml
-from scope.utils import forgiving_true, load_config, read_hdf, read_parquet
+from scope.utils import forgiving_true, load_config, read_hdf, read_parquet, write_hdf
+import json
 
 
 @contextmanager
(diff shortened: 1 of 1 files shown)


scope.py
```diff
@@ -839,6 +840,175 @@ def create_inference_script(
...
+
+        if len(np.unique(column_nums)) > 1:
+            raise ValueError('Not all predictions have the same number of columns.')
+
+        preds_df = pd.concat(df_coll, axis=0)
+        # Define non-variable class as 1 - variable
+        include_nonvar = False
+        if f'vnv_{algorithm}' in preds_df.columns:
+            include_nonvar = True
+            preds_df[f'nonvar_{algorithm}'] = np.round(
```
mcoughlin, line 912: Why the rounding?

mcoughlin (approved):
Just one question on the rounding.

scope.py
bfhealy: Currently in inference.py all classification probabilities are rounded to 2 decimal places. The same is done here to the non-variable column for consistency with the other classifications.

mcoughlin: Ah gotcha. Do you understand why you do this?

bfhealy: I think the rounding helps with readability when scanning classifications. The rounded probabilities are numerically close to the original values, but when looking at many of them together, a person can make quicker sense of the rounded values.

bfhealy:
@mcoughlin I added a few lines to fix the random state for source selection. Seems beneficial to be able to reproduce the set of selected sources.

mcoughlin:
@bfhealy good idea! merge when ready.

PR #191, opened 2022-12-05
Add initial lcstats.py from ztfperiodic

bfhealy:
This PR adds `lcstats.py` from the ztfperiodic codebase to the `tools` directory in scope. Further changes will be made after merging, but committing the initial code will help to keep track of further changes to the original.


mcoughlin (approved):
Looking forward to seeing this improved :) LGTM

PR #207, opened 2023-01-04
Consistent/custom imputation strategy

bfhealy:
This PR makes imputation consistent between training and inference. It also allows for different imputation strategies for each feature, as specified in the updated config file under `impute_strategy`. Options for imputation are `zero`, `median`, `mean` and `regress`.

The `median` and `mean` statistics are either computed from the non-NaN values of the input dataset (set `self_impute = True`) or a training set specified in the config file. The subset of features having the `regress` strategy are input to a KNN imputer that fills missing values by regression.

This PR may require a more detailed imputation approach for some features (e.g. class-by-class). For example, sources labeled as a galaxy with a missing parallax intrinsically differ from bright stars with the same missing measurement. Unlabeled sources are even more difficult to interpret because we have no initial explanation for the missing parallax.

Finally, the KNN imputation approach adds significant time to each round of inference, taking ≥ 10x longer than the previous approach.


AshishMahabal:
Looks good. 
(1) You may want to comment on why AllWISE__w3sigmpro and AllWISE__w4sigmpro are treated differently than w1 and w2.
(2) The place where you look for median. Is that replaced by mean/Mean/MEAN? Is median not being used at all now?

bfhealy:
Thanks @AshishMahabal -
(1) I added brief comment lines for the W3 and W4 mag errors explaining why we exclude them by default.
(2) Median is now being used for the features in the config file having `impute_strategy: median`. The mean is currently not being used for any features, but the code supports that statistic if the user changes the impute strategy for a given feature.

bfhealy:
Here's example output showing which technique is being used for each feature:

Screen Shot 2023-01-26 at 2 59 10 PM

AshishMahabal (approved):
Looks good.

PR #208, opened 2023-01-09
Facilitate subsequent training for existing models

bfhealy:
This PR 'closes the loop' for the Scope active learning process by:
- Updating how pre-trained models are loaded for another round of training. The default behavior is now to load the full HDF5 model that was saved during the previous round of training. The original behavior of loading only the model weights is preserved and can be performed by setting `weights_only = True`. 
- Updating the `create_training_script` method to allow pre-trained models to be specified in each call to `scope.py train`. Training will either be run on all classes listed in the config file (set `train_all = True`) or only those classes having pre-trained models in the specified group directory.
- Updating docs to reflect these changes.


doc/usage.md
```diff
@@ -78,13 +78,15 @@ for class in pnp longt i fla ew eb ea e agn bis blyr ceph dscu lpv mir puls rrly
 done;
 ```
 
-A training script can also be generated using `./scope.py create_training_script`, for example:
+If `--save` is specified during training, an HDF5 file of the model's layers and weights will be saved. This file can be directly used for additonal training and inferencing.
```
mcoughlin, line 81: additional

scope.py
```diff
@@ -587,27 +588,29 @@ def train(
...
-            optimizer=optimizer,
-            learning_rate=lr,
-            momentum=momentum,
-            monitor=monitor,
-            patience=patience,
-            callbacks=callbacks,
-            run_eagerly=run_eagerly,
-        )
+        if pre_trained_model is not None:
+            classifier.model = classifier.load(
```
mcoughlin, line 592: Any need for some kind of validation on the shape of input or similar here?

bfhealy:
@mcoughlin Good idea, I've implemented a shape check between the inputs of the dataset and the pre-trained model. If, e.g., we change which features are included between one iteration of training and the next, the code will raise a relevant error.

mcoughlin (approved):
LGTM!

Issue #125, opened 2022-10-13
tensorflow error using GPU for training on M1 Mac


bfhealy:
When running `./scope.py train` and including `--gpu 0` to specify use of the GPU, I get an error even though the GPU is recognized and available. I think this may happen because the `ResourceApplyAdamWithAmsgrad` operation is not currently supported by `tensorflow-metal` (see e.g. [this discussion](https://developer.apple.com/forums/thread/687611) and its similarity to the error messages below). I've tried upgrading to the latest version of `tensorflow-metal` (0.6.0) but still get the error. Fortunately training still runs reasonably fast on the CPU.

`tensorflow.python.framework.errors_impl.InvalidArgumentError: Cannot assign a device for operation model/conv_conv_1/separable_conv2d/ReadVariableOp: Could not satisfy explicit device specification '' because the node {{colocation_node model/conv_conv_1/separable_conv2d/ReadVariableOp}} was colocated with a group of nodes that required incompatible device '/job:localhost/replica:0/task:0/device:GPU:0'. All available devices [/job:localhost/replica:0/task:0/device:CPU:0, /job:localhost/replica:0/task:0/device:GPU:0].`

`Colocation Debug Info: 
Colocation group had the following types and supported devices: 
Root Member(assigned_device_name_index_=2 requested_device_name_='/job:localhost/replica:0/task:0/device:GPU:0' assigned_device_name_='/job:localhost/replica:0/task:0/device:GPU:0' resource_device_name_='/job:localhost/replica:0/task:0/device:GPU:0' supported_device_types_=[CPU] possible_devices_=[] 
ResourceApplyAdamWithAmsgrad: CPU
ReadVariableOp: GPU CPU 
_Arg: GPU CPU `

bfhealy:
Update: version 0.7.0 of `tensorflow-metal` does not raise the above error. However, running training on the GPU takes slightly longer per epoch (and overall) than using the CPU on my M1 Mac. 

Side note: In the latest version of `tensorflow-macos` (2.11), the Keras optimizer in `nn.py` needs to be changed to `tf.keras.optimizers.legacy.Adam` to work properly.

PR #214, opened 2023-01-18
Fix callbacks for pre-trained model, use legacy optimizer

bfhealy:
This PR fixes an error that was raised when running additional training on pre-trained models. The callbacks metadata for the pre-trained model is not saved and so cannot be loaded in like the rest of the model components. Thus, there is now a separate `set_callbacks` method in `nn.py` which is called when loading in a pre-trained model. 

Additionally, this PR adds compatibility with the latest `tensorflow-macos` (2.11), which raises an error when accessing `tf.keras.optimizers.Adam`. The replacement, `tf.keras.optimizers.legacy.Adam`, is also included in the previous `tensorflow-macos` version (2.10), so this change will not break scope on environments set up in September 2022 and onward.

(On the topic of tensorflow versions and M1 Macs, see also the updates in #125 pertaining to GPU support.)

### DIFF
--- a/.requirements/dev.txt
+++ b/.requirements/dev.txt
@@ -2,10 +2,10 @@ deepdiff>=5.0
 gsutil>=4.60
 keras-tuner>=1.0.2
 matplotlib>=3.3
-numpy<1.20
+numpy
 pytest>=6.1.2
 questionary>=1.8.1
 scikit-learn>=0.24.1
-tensorflow<2.6
+tensorflow
 tensorflow-addons>=0.12
 wandb>=0.12.1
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
Nice! LGTM!

bfhealy:
@mcoughlin I forgot that for machines other than M1 Macs, we require `tensorflow < 2.6`, explaining the test failure. I could either remove this limit on the tf version or set up a version-checking `if-else` in the `nn.py` code.

mcoughlin:
@bfhealy I think removing the limit is fine assuming everything works?

bfhealy:
@mcoughlin Sounds good, I'll remove the limit and see how the tests run.

Issue #127, opened 2022-10-13
Duplicate rows in group source downloads


bfhealy:
While this issue is connected to and raised on [SkyPortal](https://github.com/skyportal/skyportal/issues/3516), it is worth noting here as well that group source downloads from Fritz can yield files with duplicate rows. For the ~80,000 sources in the golden dataset group, ~5-10% of rows may be duplicated when using `scope_download_classification.py`. See the linked issue for more discussion about potential solutions to the problem.

bfhealy:
Closing for now, since recent downloads have not produced duplicate sources.

PR #215, opened 2023-01-19
Save obj_id for uploaded sources

bfhealy:
This PR modifies `scope_upload_classification.py` to save a version of the input file with a column of obj_ids for sources uploaded to Fritz. While in theory the obj_ids can be downloaded from Fritz, the duplicate row bug (#127) prevents the complete dataset from being obtained. Therefore it is important for the user to have a record of obj_ids produced locally after each upload. 

On uploading new information for existing sources: since one source can have multiple ZTF IDs, it is preferable to find the correct existing source using on the position-based obj_id. As such, the `no_match_ids` keyword has become `match_ids`, and when this keyword is set a new warning cautions the user about matching based on ZTF ID.

Finally, support for hdf5 and parquet format has been added.

### DIFF
--- a/tools/scope_upload_classification.py
+++ b/tools/scope_upload_classification.py
@@ -4,6 +4,7 @@ import json as JSON
 import pandas as pd
 from penquins import Kowalski
 from scope.fritz import save_newsource, api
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Very nice. LGTM

PR #216, opened 2023-01-19
Rework uploading of specific classifications

bfhealy:
Normally we set `-classification 'read'` when using `scope_upload_classification.py` to post classifications, and all classifications in the taxonomy mapper are found in the input dataset. There is additionally the ability to post specific classifications, and this PR reworks that process. The goal is to allow a dataset lacking a complete overlap of labels with the taxonomy mapper to still have classifications posted. For example, a file of YSO objects may only contain 'ra', 'dec' and 'YSO' columns with no other labels.

Previously, if N class names were specified in the `-classification` argument, the code would use the last N columns of the input dataset to source probability values, regardless of those column names. This functionality can be confusing; instead, the code now takes probabilities from columns having names in the `-classification` argument. This removes the need to ensure the correct order of columns in the input file.

Also, the `taxonomy_id` argument is no longer used and has been removed. Its functionality is included in that of the taxonomy mapper, which already specifies taxonomy ids on a class-by-class basis. The documentation has been updated accordingly.


mcoughlin (approved):
LGTM

PR #218, opened 2023-01-19
Checkpoint obj_id saves

bfhealy:
This PR checkpoints `obj_id` saving in `scope_upload_classification.py` to keep a better record of which objects have been uploaded, even if the process does not reach its completion. The checkpoint batch size is currently at 10 but may need to be increased for large input files where the repeated writes could take more time.


mcoughlin (approved):
LGTM!

PR #220, opened 2023-01-20
Handle missing photometry during source upload

bfhealy:
This PR reorganizes source uploads to skip sources without photometry. If a list of uploaded sources is requested, sources without photometry will have an empty field in the obj_id column to indicate that they were not uploaded.


mcoughlin (approved):
Nice! LGTM!

PR #222, opened 2023-01-23
Add cone search for ZTF IDs

bfhealy:
This PR adds a cone search function to `get_quad_ids.py`. This new function takes lists of obj_id, ra, and dec, queries the desired catalog, and returns the ZTF IDs of all available data within a specified radius around each set of coordinates. The function returns a dataframe containing the `_id` values from Kowalski and their associated obj_ids. If there are multiple ZTF IDs for a single obj_id, each separate ZTF ID will have its own row in the dataframe.

Upcoming development will see this function called by `scope_download_classification.py` as the first step in collecting all available light curve feature sets for each source.

### DIFF
--- a/tools/get_quad_ids.py
+++ b/tools/get_quad_ids.py
@@ -129,6 +129,94 @@ def get_ids_loop(
     return ser
 
 
+def get_cone_ids(
+    obj_id_list: list,
+    ra_list: list,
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Nifty. LGTM.

PR #223, opened 2023-01-23
Add more flexibility to get_features.py

bfhealy:
This PR updates `get_features.py` to allow more flexibility when retrieving features. The user can now specify a projection for a Kowalski query that restricts the columns returned. When `get_features.py` is called from the command line, this projection is generated from the `column_list` argument, which contains the list of features desired by the user. Furthermore, the new `features_catalog` argument allows the user to specify which catalog to query without editing the code.

The motivation for this change is the lack of ZTF filter ids in the current Scope feature collection (`ZTF_source_features_DR5`). This information may be relevant for upcoming training, so it is important to have a quick way to query a different catalog (`ZTF_sources_20210401`) for a single column (`filter`).

### DIFF
--- a/tools/get_features.py
+++ b/tools/get_features.py
@@ -224,6 +224,10 @@ def run(**kwargs):
         if True, write results and make necessary directories.
     write_csv: bool
         if True, writes results as csv file in addition to parquet.
+    column_list: list
+        List of strings for each column to return from Kowalski colleciton.
+    suffix: str
+        Suffix to add to saved feature file.
     Returns
     =======
     Stores the features in a file at the following location:
--- a/tools/get_features.py
+++ b/tools/get_features.py
@@ -48,7 +48,7 @@ def get_features_loop(
     max_sources: int = 100000,
     restart: bool = True,
     write_csv: bool = False,
-    column_list: list = [],
+    projection: list = [],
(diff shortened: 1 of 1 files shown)


tools/get_features.py
```diff
@@ -151,15 +159,17 @@ def get_features(
...
+            df_temp = df_temp.astype(dtype=dtype_dict)
         df_collection += [df_temp]
         try:
             dmdt_temp = np.expand_dims(
                 np.array([d for d in df_temp['dmdt'].values]), axis=-1
             )
         except Exception as e:
-            print("Error", e)
-            print(df_temp)
+            if "dmdt" in projection:
```
mcoughlin, line 170: Should there be an else here?

mcoughlin (approved):
Just one small question

tools/get_features.py
bfhealy: @mcoughlin I pushed a slight change to this code, but I don't think there is need for an else here. The if statement prints any dmdt-related exception if the user specifically requests dmdt or is using the default projection. Otherwise, the exception can be suppressed, since dmdt will not be among the features returned by the query.

mcoughlin: Sounds good :)

PR #226, opened 2023-01-25
Always upload batch classifications

bfhealy:
This PR moves the code that uploads batch classifications, allowing the batch to be posted even if the last source of the batch did not get uploaded.


mcoughlin (approved):
LGTM

PR #227, opened 2023-01-25
Download all available feature sets

bfhealy:
This PR primarily reworks `scope_download_classification.py` to download all available sets of features for each source on Fritz. Since multiple ZTF ids can be associated with a single source (corresponding to different light curve bands/fields), the one-to-one ZTF ID matching previously performed in this code has been removed. The new code uses the coordinate-based object ids from Fritz to connect labels and features.

A cone search finds all available feature sets within 2 arcsec (default) of the input coordinates for each Fritz source. In cases where multiple sources 'claim' a feature set, the source having the smallest Euclidean distance from the feature coordinates is assigned the set.

The ZTF filter ID for each feature set, which is currently located in a separate catalog, can be joined to the other features by setting the `-get_ztf_filters` flag.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -159,6 +159,7 @@ inputs:
 8. Name of directory to save downloaded files
 9. Name of file containing merged classifications and features
 10. Output format of saved files, if not specified in (9). Must be one of parquet, h5, or csv.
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM!

PR #229, opened 2023-01-26
Add argument to check 'labelled' box when uploading

bfhealy:
This PR adds a `-check_labelled_box` argument to `scope_upload_classification.py` which checks a source's 'labelled' box on Fritz after uploading classifications. By default, sources remain unchecked until they are reviewed during active learning.


mcoughlin (approved):
LGTM. As an aside on the version, are we planning on pypi releases and so on?

bfhealy:
I think that would be a nice goal to accomplish for this code. Not quite sure how our current hardware-dependent installation process (especially re. tensorflow) would translate to a pypi package though.

PR #230, opened 2023-01-26
Get ZTF filters when starting with Fritz downloaded file

bfhealy:
This PR passes the `get_ztf_filters` flag to the feature download code in `scope_download_classification.py` when the script's arguments include an input file from Fritz. The flag was previously missing from the function call.


mcoughlin (approved):
LGTM

PR #231, opened 2023-01-27
Add option to delete existing classifications during upload

bfhealy:
This PR allows the user to delete/replace each source's existing classifications when uploading new classifications. This will be useful for active learning purposes as we replace outdated classifications from a previous round of training.


mcoughlin (approved):
LGTM

PR #232, opened 2023-01-27
Update training pipeline

bfhealy:
This PR updates the training pipeline in two ways:
- Any `numpy.float64` columns in the features datafame are converted to `numpy.float32` to be compatible with Tensorflow. This behavior is customizable with the `float_convert_types` argument.
- Feature stats are no longer sourced from the values in the config file by default. Instead, they are computed after imputation by a new function in `utils.py`. Setting `feature_stats = "config"` will produce the old default behavior.


mcoughlin (approved):
Very nice.

Issue #233, opened 2023-01-27
Imputing features for inference sources


bfhealy:
When imputing features for inference sources, we can either:
(1) always use the mean/median/KNN regression results for sources in the training dataset, or
(2) self-impute using stats computed from each batch of inference sources (usually 100,000 sources in a batch). 

Option (1) has the advantage of a uniform approach for all inference fields/batches, but if there are significant discrepancies between the training and inference source populations, the imputed values may be inaccurate. Option (2) tailors imputation to the characteristics of each individual batch of sources at the cost of a less uniform approach. Our code can already support both approaches - which do you think is preferable @AshishMahabal, @mcoughlin?

(Regardless of the choice above, all features will be scaled using identical statistics to ensure consistency between training and inference).

mcoughlin:
(2) feels better, especially as I assume the batches will be in the same field and therefore similar parts of the sky.

AshishMahabal:
I agree.

bfhealy:
Thank you both!

PR #234, opened 2023-01-27
Inference updates

bfhealy:
This PR updates the inference pipeline as follows:
- Features are self-imputed based on each batch of sources (see #233).
- The sequence of imputation -> feature stats -> scaling is now the same as the training process.
- Feature stats are by default generated with the new function in `utils.py` (see #232). The stats are generated from the training set to ensure scaling consistency between training and inference.
- The `create_inference_script` method has a new argument allowing the user to specify the feature scaling method.


mcoughlin (approved):
NICE! LGTM.

bfhealy:
Thanks!

PR #237, opened 2023-01-30
Add more subtlety to classification replacement

bfhealy:
This PR adds more subtlety to classification replacement in `scope_upload_classification.py` by only deleting all existing classifications if they are all included in at least one of the user's target groups. This is expected to usually be the case, but if a source already existed in another group with its own separate classifications, these classifications should be preserved even if the user is otherwise replacing their own classifications.


mcoughlin (approved):
LGTM

PR #239, opened 2023-02-01
Consolidate inference results

bfhealy:
This PR adds code to consolidate inference results such that there is only one row per source. The code supports taking the mean, median or max of multiple classification probabilities to compute the consolidated value. Unique sources are identified using each row's associated Gaia EDR3, AllWISE or Pan-STARRS ID (in that order). In addition to the AL subset which gets produced, the new code can also save the full list of all rows and the full consolidated source list.

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -927,14 +927,27 @@ class Scope:
         statistic: str = 'mean',
     ):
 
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM.

PR #240, opened 2023-02-02
Update active learning pipeline

bfhealy:
This PR updates the active learning pipeline in several ways. The primary update is to add a new method of selecting the AL subset from the multi-field consolidated sample. Since some classifications do not have any examples with probabilities above a certain high-confidence value (e.g. 0.9), this new method instead selects the top N sources for each class. The user can still specify the `probability_threshold` below which classifications are ignored, but the threshold can now be reduced and still yield only the most confident sources for each class. This new functionality can be called by setting the `--select_top_n` keyword in the call to `select_al_sample`.

There is also the option, enabled by setting `--include_all_highprob_labels`, to include all classifications above `probability_threshold` for the sources identified by the top N selection. This can provide a more complete set of classifications without going too far: For example, if a source has a top N `variable` label and also has `periodic` probability = 0.9, `periodic` should not necessarily be left out just because it is not in the top N for its class.

Additionally:
- Since sources can appear in multiple fields, predictions are consolidated one more time (after consolidating each field)
- The generation of the position-based `obj_id` column takes place after this final consolidation
- The period column is aggregated along with predictions and coordinates
- The consolidated file, which can take a few hours to generate for millions of sources (currently lots of pandas DataFrame manipulation), can be loaded in to the AL code to save time when using the same file to select a different subset of features

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -1125,7 +1125,8 @@ class Scope:
 
         :return:
 
-        :example:  ./scope.py select_al_sample --fields=[296,297] --group='experiment' --min_class_examples=1000 --probability_threshold=0.9 --exclude_training_sources --write_consolidation_results
+        :examples:  ./scope.py select_al_sample --fields=[296,297] --group='experiment' --min_class_examples=1000 --probability_threshold=0.9 --exclude_training_sources --write_consolidation_results
+                    ./scope.py select_al_sample --fields=[296,297] --group='experiment' --min_class_examples=500 --select_top_n --probability_threshold=0.7 --exclude_training_sources --read_consolidation_results
         """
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM. At some point, would be good to get tests for some of this stuff going..

bfhealy:
@mcoughlin Agreed, the active learning pipeline should definitely be added to the tests, and it should be pretty straightforward to build on the current tests to do that. It's a bit more involved for the upload/download classifications code, since testing that requires either firing up SkyPortal or hitting Fritz with API calls.

mcoughlin:
@bfhealy makes sense, good to be reasonable here.

PR #241, opened 2023-02-02
Update active learning mapper with new taxonomy IDs

bfhealy:
This PR updates the mapper for local active learning classifications and the latest Fritz taxonomy IDs.


mcoughlin (approved):
LGTM

PR #242, opened 2023-02-03
Allow classifications belonging to no groups to be replaced

bfhealy:
This PR allows the replacement of classifications belonging to no groups on Fritz. Currently `scope_upload_classification.py` will not identify a classification for deletion if its `groups` field is empty. In the case that sources in our current active learning group on Fritz overlap with our new sample and need their classifications replaced, this change will prevent the old classifications from remaining along with the new ones.


mcoughlin (approved):
Maybe leave a comment on what -1 is doing exactly.

bfhealy:
@mcoughlin Good idea!

PR #243, opened 2023-02-03
Post new sources using existing obj_ids and rework source annotations

bfhealy:
This PR makes additional changes to the classification upload process:

- Our active learning code generates `obj_id` values for all sources, and it is preferable to use these ids when posting sources (rather than have the code generate them at the time of upload). Since these AL sources are not yet posted to Fritz, if `-use_existing_obj_id` is set, the current `fritz.py` code would skip posting each source. Thus, the code is changed to treat these sources as new, even though `obj_id` is provided. Re-posting sources that already exist on Fritz is still avoided.

- The other major change in this PR is the restructuring of the annotation upload process, such that the `survey_id` used to identify unique sources (one of Gaia EDR3, AllWISE or PS1 IDs) can be posted with a 'SCoPe' origin instead of the `ztf_id` which is no longer used to identify unique sources.


tools/scope_upload_classification.py
```diff
@@ -177,17 +175,16 @@ def upload_classification(
...
-                        warnings.warn(
-                            "One source may have more than one ZTF ID. It is recommended to match to an exising source's obj_id by setting -use_existing_obj_id."
-                        )
+                        print("Attempting to match with survey_id annotation.")
                         annotations = src['annotations']
                         for annot in annotations:
-                            if annot['origin'] == ztf_origin:
-                                src_ztf_ids = [int(x) for x in annot['data'].values()]
-                                if ztfid in src_ztf_ids:
+                            if annot['origin'] == 'SCoPe':
```
mcoughlin, line 181: Don't we want this tunable? Or don't we care?

mcoughlin (approved):
LGTM. Small comment.

tools/scope_upload_classification.py
bfhealy: @mcoughlin Probability a good idea to make it tunable; just made the change

PR #247, opened 2023-02-07
Make checkbox optional when deleting classifications

bfhealy:
This PR extends the functionality of the `-check_labelled_box` keyword in `scope_upload_classification.py` to classification deletion when `-replace_classifications` is set. This fixes an issue where sources would be checked as labelled if their classifications were replaced by the upload script.

(Anticipates deploy of the related SkyPortal change to Fritz)


mcoughlin (approved):
LGTM

PR #248, opened 2023-02-08
Add feature imputation to downloads and fix filter bug

bfhealy:
This PR adds the option to impute missing features in `scope_download_classification.py`. It also fixes a bug that was assigning the wrong ZTF filter IDs to downloaded sources. This bug has not affected current active learning efforts.


mcoughlin (approved):
LGTM!

PR #249, opened 2023-02-13
Save downloaded sources from Fritz before getting features

bfhealy:
This PR modifies `scope_download_classification.py` to save an additional file containing each source in the specified Fritz group. This file is useful to save before the feature merging process begins, since that process can return 0, 1 or >1 rows for each source on Fritz.


mcoughlin (approved):
LGTM

PR #250, opened 2023-02-13
Rename galaxy to extended in golden dataset mapper

bfhealy:
All training sources on Fritz having the `galaxy` label now also have the `extended` label from the latest SCoPe Phenomenological Taxonomy. Thus, this PR renames `galaxy` to `extended` in both the golden dataset label mapper and the default config file. In the config file, two other classes with few positive examples also have a new `batch_size` argument in their hyperparameter info.


mcoughlin (approved):
LGTM

PR #252, opened 2023-02-14
Skip Kowalski filtering by n points if minobs is zero

bfhealy:
This PR allows the `get_field_ids` function within `get_quad_ids.py` to return ids for `ZTF_sources_...` catalogs, which do not contain the `n` entry created during feature generation. The `{"n" : {"$gt": minobs}}` filter parameter is now only included if `minobs` is nonzero. `get_field_ids` can now be used at the start of the feature generation process in addition to the inference process.


mcoughlin (approved):
LGTM

PR #253, opened 2023-02-14
Make saving ids optional in get_ids_loop

bfhealy:
This PR makes saving optional in `get_ids_loop`. The default remains to save the ids. The code now also prints the total number of IDs found.


mcoughlin (approved):
LGTM!

PR #254, opened 2023-02-15
Split get_lightcurves into two functions

bfhealy:
This PR splits the `get_lightcurves` function in `fritz.py` into two functions: `get_lightcurves_via_coords`, which replicates the original functionality and calls `get_lightcurves_via_ids`, which queries Kowalski based on the `_id` field directly. This change allows the user to get light curves from a list of ids without the need for a coordinate-based query, which will be useful for feature generation.

### DIFF
--- a/scope/fritz.py
+++ b/scope/fritz.py
@@ -133,7 +133,9 @@ def radec_to_iau_name(ra: float, dec: float, prefix: str = "ZTFJ"):
     return prefix + hms + dms
 
 
-def get_lightcurves(gloria, ra, dec, radius=2.0):
+def get_lightcurves_via_coords(
+    gloria, ra, dec, radius=2.0, catalog=None, program_id_selector=list([1, 2, 3])
+):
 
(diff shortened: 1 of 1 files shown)


scope/fritz.py
```diff
@@ -142,11 +144,9 @@ def get_lightcurves(gloria, ra, dec, radius=2.0):
...
     ]
 
-    # simply select last catalog
-    catalog = available_catalogs[-1]
-
-    # allow access to all data by default
-    program_id_selector = list([1, 2, 3])
+    if catalog is None:
+        # simply select last catalog
+        catalog = available_catalogs[-1]
```
mcoughlin, line 149: Maybe print or log what is used then?

mcoughlin (approved):
Small comment. Looks good.

scope/fritz.py
bfhealy: Agreed, just pushed that change.

PR #255, opened 2023-02-16
Add initial periodsearch from ztfperiodic

bfhealy:
This PR adds the periodsearch code from ztfperiodic to scope/tools. The code has been cleaned of unused variables/imports and a ztfperiodic utils dependency has been added to scope/utils.py.


mcoughlin (approved):
Ugly ;) Let's plan on cleaning it up when we can.

PR #256, opened 2023-02-16
Pin periodfind

bfhealy:
This PR pins [periodfind](https://github.com/ejaszewski/periodfind) (CUDA-accelerated period-finding algorithms) to the scope repo in an effort to have as much relevant code in one place as possible.


mcoughlin (approved):
LGTM

PR #258, opened 2023-02-17
Fix failing documentation generation

bfhealy:
This PR attempts to fix failing documentation generation by updating requirements.


mcoughlin (approved):
LGTM! I also wonder if we can start deploying the docs to scope.github.io or similar?!

bfhealy:
That sounds like a good idea! I'll look into what that involves.

PR #262, opened 2023-02-21
Add dmdt feature code to lcstats

bfhealy:
This PR adds the `compute_dmdt` function (adapted from https://github.com/dmitryduev/kowalski-archived/blob/7b9582fae6cd37bbd334bca228ef429d96e0e498/kowalski/dev/ingest_ZTF_source_features.py#L439) to `lcstats.py`. A `numba` dependency is also added to the scope requirements to satisfy behind-the-scenes code for this feature.

The default dmdt bin intervals have been changed from `v20200318` to  `v20200205`, allowing for longer-timescale variability to be distinguished on the dmdt histograms. See the added code below for the exact bin intervals.


AshishMahabal (changes requested):
I don't seem to see the latest dmdt version. What is it called? It should be something like v202302??, right?
Also, we had removed 1/145,2/145,3/145 days that had been originally used with CRTS because of our cadence. However, if we wish to take advantage of the fast cadence where we sit at one location, we may want to introduce a couple bins at that end. When there is no high-cadence data those set of bins will be empty, but will not hurt the classification.

bfhealy:
@AshishMahabal The current default for the bins is `v20200205` - I didn't rename it yet from the date that was in the code. I will take a look at adding some of those faster cadence bins again and create a new dictionary entry with something like `v20230221`. Should we still maintain a 26x26 shape for each histogram?

bfhealy:
@AshishMahabal I should also mention that following the recipe of the first Scope paper, before generating features I plan to drop all but the first datapoint in a group of points separated by less than 30 mins. Thus there may be limited use for the short-cadence bins right now.

bfhealy:
Hi @AshishMahabal, please let me know if the intervals under `v20230221` are appropriate. They would maintain a 26x26 histogram, use the same magnitude bins as set for DR5, and extend the time baseline out to 2000 days.

AshishMahabal (approved):
Looks good. We should consider using the high cadence sequences at least separately. May be a dmdt just when there is high cadence data.

tools/featureGeneration/lcstats.py
```diff
@@ -16,9 +16,252 @@
 from scipy.optimize import curve_fit
 from scipy.signal import sawtooth
 
+from numba import jit
+
 warnings.filterwarnings("ignore")
 
 
+# Define bin intervals for dmdt histograms
```
mcoughlin, line 24: Is this better in the config? Or do you prefer it in the code?

mcoughlin (approved):
Just a question if you want the bins in the config instead.

tools/featureGeneration/lcstats.py
bfhealy: Seems better to put it in the config now that you mention it.

PR #264, opened 2023-02-22
Return organized list of ccd/quadrant ids

bfhealy:
This PR modifies `get_ids_loop` in `get_quad_ids.py` to return an additional variable containing a list of each CCD/quadrant's ZTF IDs. The list is organized such that each element contains all the IDs for one CDD/quadrant pair.


mcoughlin (approved):
LGTM

PR #265, opened 2023-02-22
Add util to remove high-cadence data

bfhealy:
This PR adds a function to `scope/utils.py` to remove all but the first high-cadence data points from a light curve.

### DIFF
--- a/scope/utils.py
+++ b/scope/utils.py
@@ -637,7 +637,7 @@ def overlapping_histogram(a, bins):
 
 def removeHighCadence(t, m, e, cadence_minutes=30.0):
     idx = []
-    for ii, _ in enumerate(t):
+    for ii in range(len(t)):
         if ii == 0:
             idx.append(ii)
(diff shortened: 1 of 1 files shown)


scope/utils.py
```diff
@@ -635,6 +635,20 @@ def overlapping_histogram(a, bins):
     return n, (bins[:, 0] + bins[:, 1]) / 2.0
 
 
+def removeHighCadence(t, m, e, cadence_minutes=30.0):
+    idx = []
+    for ii, _ in enumerate(t):
```
mcoughlin, line 640: Probably don't need enumerate here.

mcoughlin (approved):
Small comment.

scope/utils.py
bfhealy: Agreed and replaced.

mcoughlin: Should be good to merge!

PR #266, opened 2023-02-22
Add option to get each object's geojson coordinates along with ID

bfhealy:
This PR adds a `get_coords` keyword to `get_ids_loop`. Setting this keyword returns a list of dictionaries for each source structured as  `{_id : radec_geojson}`. The quadrant-by-quadrant list structure is still maintained.

This update will allow for easier crossmatching with non-ZTF catalogs on Kowalski before fetching the light curves (e.g. to remove sources close to bright stars).


mcoughlin (approved):
LGTM!

PR #267, opened 2023-02-23
Add bright star exclusion radius utils

bfhealy:
This PR adds utils to:
- Convert Gaia $G < 13$ and $BP-RP$ magnitudes to Tycho $B$ and $V$
- Calculate the radius within which to exclude sources near bright stars (as a function of $B$ and $V$).


mcoughlin (approved):
Thanks @bfhealy

PR #268, opened 2023-02-23
Rename gloria to kowalski_instance

bfhealy:
This PR renames the `gloria` variable to `kowalski_instance` in `get_quad_ids.py` and `fritz.py`. For functions in `get_quad_ids.py`, the instance can also now be passed in as an argument. These changes assist upcoming feature generation code, which requires querying multiple instances of Kowalski machines.


mcoughlin (approved):
LGTM!

PR #269, opened 2023-02-24
Add Kowalski timeout to config, update collections

bfhealy:
This PR adds a Kowalski timeout to the config file and updates source/feature collections to the most recent available versions.


mcoughlin (approved):
We should note that these do not match (i.e. what DR the sources are)..

bfhealy:
Agreed, I added some comments about that in the file.

PR #270, opened 2023-02-24
Allow single ccd/quad to be passed into get_ids_loop

bfhealy:
This PR allows ints to be passed to the `ccd_range` and `quad_range` parameters in `get_ids_loop`. Previously a list of two ints had to be passed in even if the user only wanted one ccd/quadrant done.


mcoughlin (approved):
LGTM

PR #271, opened 2023-02-24
Add feature generation script

bfhealy:
This PR adds a preliminary script in `tools/generate_features.py` to generate features from ZTF light curves. Upon running `./generate_features.py`, the script performs the following:

- Get IDs/coords for a field/ccd/quad combo
- Query Gaia for nearby sources brighter than G=13 (default 300 arcsec radius, ~the largest exclusion radius of A. Drake's findings)
- Calculate exclusion radii and drop sources too close to a bright star
- Get light curves for remaining sources
- Drop all flagged light curve points
- Drop all but the first point from groups of high-cadence data (default 30 min)
- Apply a threshold for minimum number of points (default 50) after the above cuts
- Compute 'features': currently a few test statistics (to be replaced by lcstats, periodfind etc.)
- Write results in parquet format to `scope/generated_features` directory (directory/filenames can be customized)

Commented code is intended to be implemented in a future PR.

### DIFF
--- a/scope/__init__.py
+++ b/scope/__init__.py
@@ -5,7 +5,7 @@ from .fritz import *
 
 # Below code adapted from https://github.com/skyportal/skyportal/blob/main/skyportal/__init__.py
 # 2022-10-18
-__version__ = '0.4.dev0'
+__version__ = '0.5.dev0'
 
 if 'dev' in __version__:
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
Nice.

PR #273, opened 2023-02-24
Get ZTF alert stats

bfhealy:
This PR adds code to get the `n_ztf_alerts` and `mean_ztf_alert_braai` features from the `ZTF_alerts` catalog on `kowalski`. The new code in `tools/featureGeneration/alertstats.py` is called by `generate_features.py`.

Since my `kowalski` token is different than the one that covers `gloria` and `melman`, I added an `alt_token:` field in `config.defaults.yaml` to allow me to authenticate a `kowalski` instance for the alerts. In this file there is also a new `alerts` field that allows the user to specify the name of the alerts collection.

### DIFF
--- a/config.defaults.yaml
+++ b/config.defaults.yaml
@@ -3,6 +3,7 @@ kowalski:
   port: 443
   protocol: https
   token:
+  alt_token:
   timeout: 300
   collections:
     gaia: Gaia_EDR3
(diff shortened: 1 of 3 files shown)


config.defaults.yaml
```diff
@@ -3,6 +3,7 @@ kowalski:
   port: 443
   protocol: https
   token:
+  alt_token:
```
mcoughlin, line 6: Maybe document what this corresponds to?

mcoughlin (approved):
I just suggest clarifying the token stuff in the config.

config.defaults.yaml
bfhealy: Good idea, thanks!

PR #274, opened 2023-02-24
Add basic light curve stats to feature generation

bfhealy:
This PR adds basic light curve stats to `generate_features.py` by calling `lcstats.calc_basic_stats()`. It replaces the placeholder light curve features in #271.


mcoughlin (approved):
Good progress

PR #275, opened 2023-02-25
Fix failing doc tests

bfhealy:
This PR fixes the failing doc tests seen after today's PRs have been merged. The source of the failures is the updated config-default source catalog (ZTF DR15, `ZTF_sources_20230109`), which is only available on `melman.caltech.edu`. Since `gloria` is still relied upon for many scripts in scope, it would be disruptive to change the default host to `melman`. Instead, this PR comments out the line providing the config-default catalog in `scope.py doc`. The code now defaults to the `ZTF_sources_20210401` (DR5) specified in the `catalog` argument for `_get_light_curve_data()`.

In the future, hopefully this line can be uncommented and a `KowalskiInstances` class will manage the various instances more seamlessly.

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -265,8 +265,6 @@ class Scope:
         }
         response = self.kowalski.query(query=query)
         light_curves_raw = response.get("data").get(catalog).get("target")
-        print(light_curves_raw)
-        print(len(light_curves_raw))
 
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Good to keep this working.

PR #276, opened 2023-02-27
Add period-finding and Fourier stats to feature generation

bfhealy:
This PR adds period finding using `periodsearch.py` and Fourier stats using `lcstats.py` to the feature generation script. The period finding is run outside of the source-by-source loop to take advantage of GPU-accelerated algorithms that run on batches of light curves.

### DIFF
--- a/tools/generate_features.py
+++ b/tools/generate_features.py
@@ -18,10 +18,10 @@ import pandas as pd
 from astropy.coordinates import SkyCoord
 import astropy.units as u
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
@bfhealy would it be hard to add a test for some basic feature calc? Not for the GPU stuff for now.

bfhealy:
@mcoughlin I think it will be straightforward enough to add a test for this script on some sample data. I'll take a look at that.

PR #277, opened 2023-02-27
Add feature generation test

bfhealy:
This PR adds a test of our feature generation script to `scope.py test`. It runs successfully on my computer but may encounter issues on GH Actions because of the secret Kowalski token. Locally I need to use two tokens to access all of `kowalski`, `gloria` and `melman`. Currently the test script attempts to use the same token to authenticate all machines. If that fails, we can add another token to the repo's secrets.

This PR also includes a bug fix for when the number of sources in a field/ccd/quad is an integer multiple of the query size limit.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,7 +3,7 @@ name: Build, test, and upload docs
 on:
   push:
     branches: [ main ]
-  pull_request_target:
+  pull_request:
     branches: [ main ]
 jobs:
   approve: # First step
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - main
   pull_request:
+    branches:
+      - main
 jobs:
   docs:
     runs-on: ubuntu-latest
@@ -31,9 +33,8 @@ jobs:
         ./scope.py lint
     - name: Build docs
       env:
-        DUMMY_TOKEN: ${{ secrets.DUMMY_TOKEN }}
-        #KOWALSKI_TOKEN: ${{ secrets.KOWALSKI_TOKEN }}
-        #KOWALSKI_ALT_TOKEN: ${{ secrets.KOWALSKI_ALT_TOKEN }}
+        KOWALSKI_TOKEN: ${{ secrets.KOWALSKI_TOKEN }}
+        KOWALSKI_ALT_TOKEN: ${{ secrets.KOWALSKI_ALT_TOKEN }}
       run: |
         ./scope.py doc
     - name: Install SSH Client 🔑
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,8 +31,9 @@ jobs:
         ./scope.py lint
     - name: Build docs
       env:
-        KOWALSKI_TOKEN: ${{ secrets.KOWALSKI_TOKEN }}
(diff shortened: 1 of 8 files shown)


bfhealy:
@mcoughlin It seems that when running the test suite with GH Actions, the `KOWALSKI_TOKEN` env variable is not accessible within functions that I call in `scope.py`.  Is it secure to write the token to the config file within the test run so it can then be accessed by `generate_features.py` and `get_quad_ids.py`?

mcoughlin:
@bfhealy that won't work if it ends up in the artifacts of the actions. Maybe do a check with some dummy values first?

bfhealy:
@mcoughlin Good idea, thanks.

bfhealy:
@mcoughlin None of the tests generated artifacts when I wrote the dummy token to the config file. I think it's safe to use this approach to pass along the token to the other functions.

mcoughlin:
@bfhealy Ok!

bfhealy:
@mcoughlin I'm starting to think the kowalski token currently stored as a GitHub secret is no longer authenticating. Until this PR, the only test that used the token was the docs test, and upon checking the logs I see that docs has only been passing because it returns after printing "Kowalski connection not established, cannot generate docs." I could try replacing the secret token with my own.

mcoughlin:
@bfhealy That sounds fine.

bfhealy:
@mcoughlin Ok, will do!

bfhealy:
@mcoughlin From my latest tests I think I've narrowed down the issue to GH Actions not passing Kowalski secrets to the runner when the action is triggered from a forked repo (see below). This would also explain why the `docs` test in past PRs does not connect to Kowalski, but then successfully does upon merging the PR. The test still passed in all situations because of a `return` in the event that Kowalski failed to authenticate. 

Screen Shot 2023-02-28 at 11 01 55 AM

It's tough to test the query-heavy feature generation script without Kowalski access. I think we could address this by using `on: pull_request_target` instead of `on: pull_request` in the workflow files (https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/). PR-prompted tests would then be granted access to the necessary secrets, but the code/workflows for the tests would be sourced from the target repo rather than the incoming code to be merged. If PRs from first-time contributors continue to require approval to run workflows, I think we could protect Kowalski while allowing the tests the access they need.

mcoughlin:
@bfhealy Ah yes that makes sense. Your intuition seems to make sense.

bfhealy:
@mcoughlin The easiest way to get `pull_request_target` to work as intended with Kowalski tokens is to checkout the PR branch before running the workflow jobs. But this is also insecure because once any user merges their first PR (which requires approval to run workflows), they will have continued access to Kowalski/other secrets via subsequent PRs.

I've been following this article (https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci) to use a Github environment and its protection rules to gate-keep actions that involve Kowalski tokens. The downside of this technique is that it requires two manual clearances for each PR/push (one to run `test` and one for `docs`). We can authorize up to 6 people to provide clearance via an email prompt (see below), so for now it wouldn't be an issue for trusted contributors to self-clear their PRs to run tests.

If we decide to go the route of manual clearances, I'll open another PR to make the change. We'd need the updated workflows to be in place before `pull_request_target` works successfully on this PR to enable further testing of `scope.py test`.

Screen Shot 2023-02-28 at 1 35 45 PM

mcoughlin:
@bfhealy We can give it a try I think.

PR #281, opened 2023-02-28
Add new feature generation test

bfhealy:
This PR adds a test of our feature generation script to `scope.py test`. The PR also includes a bug fix for when the number of sources in a field/ccd/quad is an integer multiple of the query size limit.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -54,7 +54,7 @@ jobs:
 
     - name: Deploy docs
       if: github.ref == 'refs/heads/main'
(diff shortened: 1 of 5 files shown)


bfhealy:
@mcoughlin All tests are passing! I'm a little concerned about the failing docs deployment before I bypassed it just now, but that failure could have come from trying to run it in the wrong situation (PR instead of push).

If you'd like, I can add you to the trusted users who receive GH action clearance requests for future PRs. It will add two emails per PR/commit to your inbox though. I'm also happy to serve as the gatekeeper myself for the time being.

mcoughlin (approved):
Awesome. LGTM!

PR #282, opened 2023-02-28
Change condition for deploying docs

bfhealy:
The new tests are passing except for a failed docs deployment. The docs workflow should not even be attempting to deploy until a PR is merged, so something is not working as it used to. This PR changes the condition for deploying docs within Github Actions.

I think the change from `pull_request` to `pull_request_target` sets `github.ref` to `'refs/head/main'` even within a PR branch. This triggers a failed deployment attempt even though there is no merge happening. The change to `if: github.event_name == 'push'` should fix this, but that will not be reflected in the tests for this PR.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -47,14 +47,14 @@ jobs:
       run: |
         ./scope.py doc
     - name: Install SSH Client 🔑
-      if: github.ref == 'refs/heads/main'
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
@bfhealy Feel free to merge as much as necessary to test this out and such.

bfhealy:
@mcoughlin Great, thanks!

PR #283, opened 2023-03-01
Add cone search queries for external catalog features

bfhealy:
This PR adds external catalog crossmatches via cone search to the feature generation script. It adds a new section of the config file to specify which external features to get, then passes that info to `tools/featureGeneration/external_xmatch.py` which runs batch queries. The external features are incorporated into the existing set produced by `generate_features.py`.


mcoughlin (approved):
Looks good!

PR #284, opened 2023-03-02
Pin scope-phenomenology

bfhealy:
This PR pins the [scope-phenomenology](https://github.com/bfhealy/scope-phenomenology) repo, which contains the SCoPe Phenomenological Taxonomy used on Fritz. This removes the need to clone the repo separately from scope.


mcoughlin (approved):
LGTM

PR #285, opened 2023-03-02
Only deploy docs if merging to ZTF-owned scope repo

bfhealy:
This PR modifies the `docs` workflow to only attempt documentation deployment for pushes to the ZTF-owned scope repo (this one).

This should change nothing about actions behavior in this repo, but it should fix failing tests when pushing to forked repos and protect the docs from unintended deployments.


mcoughlin (approved):
Nice idea.

PR #286, opened 2023-03-02
Choose nearest of multiple crossmatch sources

bfhealy:
This PR address cases during feature generation where multiple external catalog sources are returned in a cone search based on a single input source. In these cases, the external catalog source having the smallest spatial separation from the input is selected.


mcoughlin (approved):
LGTM

bfhealy:
@mcoughlin Looks like GH Actions is having some issues right now. I'll hold off on merging for the time being.

PR #287, opened 2023-03-03
Add option to generate cesium features

bfhealy:
This PR allows users to generate cesium features in addition to the default scope/ztfperiodic features. The config file now contains the full list of available cesium features, and uncommented features will be generated if `-doCesium` is set in `generate_features.py`. 

Some overlapping features between scope/cesium and all periodic features are commented by default - the latter group because it requires an additional Lomb-Scargle analysis. Going forward we could modify cesium to accept a period from our existing analysis and compute Fourier features based on that value.

### DIFF
--- a/tools/generate_features.py
+++ b/tools/generate_features.py
@@ -20,11 +20,12 @@ import astropy.units as u
 from datetime import datetime
 from tools.featureGeneration import lcstats, periodsearch, alertstats, external_xmatch
 import warnings
+from cesium.featurize import time_series, featurize_single_ts
+
 
 # import time
 # import periodfind
 # from numba import jit
-# from cesium.featurize import time_series, featurize_single_ts, featurize_time_series, featurize_ts_files
 # import cesium.features as fts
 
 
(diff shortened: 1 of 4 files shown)


tools/generate_features.py
```diff
@@ -220,7 +221,7 @@ def drop_close_bright_stars(
 # def generate_features(source_catalog=source_catalog, alerts_catalog=alerts_catalog, gaia_catalog=gaia_catalog, bright_star_query_radius_arcsec=300.,
 # xmatch_radius_arcsec=2., kowalski_instances = kowalski_instances, limit=10000, period_algorithm='LS', period_batch_size=1, doCPU=False, doGPU=False, samples_per_peak=10, doLongPeriod=False, doRemoveTerrestrial=False,
 # doParallel=False, Ncore=8, doAllFields=False, field=296, doAllCCDs=False, ccd=1, doAllQuads=False, quad=1, min_n_lc_points=50, min_cadence_minutes=30., dirname='generated_features',
-# filename='features', doNotSave=False, stop_early=False):
+# filename='features', doCesium=False, doNotSave=False, stop_early=False):
```
mcoughlin, line 224: What are these commented bits?

mcoughlin (approved):
small question

tools/generate_features.py
bfhealy: I've been using those lines to keep track of potential future additions to the script (e.g. `doAllFields`, `doAllCCDs` and `doAllQuads` flags). At this point those flags might be unnecessary since the field/ccd/quad specifications will likely be incorporated into the slurm script, so I'm ok with removing those comments.

mcoughlin: up to you!

PR #288, opened 2023-03-06
Generate slurm scripts

bfhealy:
This PR adds code to generate slurm scripts to run scope feature generation in parallel on HPC resources. Changes include:
- `generate_features.py` (modified):
  - Syntax update: all keywords now begin with `--` instead of `-`
  - New check for cases where no light curves pass selection criteria
  - New keywords have been added: `--doQuadrantFile`, `--quadrant_file` and `--quadrant_index` allow the field/ccd/quadrant to be specified using a generated `slurm.dat` file (described below)

- `generate_features_slurm.py` (new):
  -  Optionally checks ZTF fields/ccd/quadrants for sources
  - Maps each field/ccd/quadrant combo with sources to a quadrant index and saves a `slurm.dat` file
  - Saves `slurm.sub` file with `SBATCH` commands and a call to `generate_features.py`. This call includes `--quadrant_index $PBS_ARRAYID`, which gets replaced with the appropriate index for each run
  - Saves `slurm_submission.sub` with `SBATCH` commands and a call to `generate_features_job_submission.py` (see below)

- `generate_features_job_submission.py` (new):
  - Calls multiple runs of `generate_features.py` for different quadrant indices
  - By default, caps number of runs at 100 and checks if any jobs are completed every 10 minutes
  - Initiates new jobs until all are completed

### DIFF
--- a/tools/generate_features.py
+++ b/tools/generate_features.py
@@ -23,11 +23,6 @@ import warnings
 from cesium.featurize import time_series, featurize_single_ts
 
 
-# import time
-# import periodfind
-# from numba import jit
-
-
 BASE_DIR = pathlib.Path(__file__).parent.parent.absolute()
 
 # setup connection to Kowalski instances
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
Looks good @bfhealy

PR #290, opened 2023-03-08
Add get_quad_ids and get_features tests

bfhealy:
This PR adds new tests for functions in `get_quad_ids.py` and `get_features.py`. The broader intent is to link these up with future tests for `inference.py` and the active learning code in `scope.py`.

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -1410,7 +1410,7 @@ class Scope:
         :return:
         """
         import uuid
-        from tools import generate_features
+        from tools import generate_features, get_quad_ids, get_features
 
         # Test feature generation
         test_field, test_ccd, test_quad = 297, 2, 2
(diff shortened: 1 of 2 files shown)


bfhealy:
Note to self to add `os.environ.get("KOWALSKI_TOKEN")` and such to `get_features` so the test passes.

scope.py
```diff
@@ -1439,6 +1439,37 @@ def test(self):
         )
         os.remove(path_gen_features)
 
+        print("Testing get_cone_ids...")
+        id_df = get_quad_ids.get_cone_ids(
+            obj_id_list=['obj1', 'obj2', 'obj3'],
+            ra_list=[40.0, 41.0, 42.0],
+            dec_list=[50.0, 51.0, 52.0],
+        )
+        print(id_df)
```
mcoughlin, line 1448: Needed?

mcoughlin (approved):
Can presumably delete the print. Otherwise LGTM

scope.py
bfhealy: Deleted, thanks!

PR #291, opened 2023-03-08
Add tests for inference and active learning sample selection

bfhealy:
This PR adds tests for `inference.py` and `scope.py select_al_sample`.


mcoughlin (approved):
LGTM! Thanks for adding tests!

Issue #236, opened 2023-01-30
Minimize imputation calls for inference


bfhealy:
Same core issue as #235, but with different details for the inference process:

- During the feature download process for inference fields (`tools/get_features.py`), the code could offer the option of imputing features and saving a separate file containing those values. This would avoid the current behavior of running `impute_features` for every batch and class during inference.

PR #292, opened 2023-03-08
Impute features when using get_features.py

### LINKED ISSUE #235: Minimize imputation calls for training
Our new imputation strategy can add significant time to training/inference steps if called too many times. We should find ways to minimize these calls even if means adding an additional step to the process. For training:

- Currently the training code calls `utils.impute_features` when each classifier is trained. We can instead impute features on the training set once, save that file and use it as input for all classifiers. This should be specified in the documentation and perhaps added as a method in `scope.py` for further ease of use.

bfhealy:
This PR adds the option to impute features when calling `get_features.py` using the strategies listed in the config. By using this new option, the user can both query features from Kowalski and impute missing features at the same time.

After imputation, the code drops rows that still have missing features if they have `include: True` in the config. That is, if there is no imputation strategy for a missing feature but that feature is included in training/inference, the row will be dropped to avoid downstream errors. For our current training set with ~180,000 light curves, only one source gets dropped.

This PR (along with #248) resolve #235 and #236 by minimizing the amount of imputation performed before training and inference.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -95,7 +95,7 @@ If `--pre_trained_group_name` is specified and the `--train_all` keyword is set,
 * The commands for inference of the field `<field_number>` (in order)
   ```
   ./tools/get_quad_ids.py --field=<field_number> --whole-field
-  ./tools/get_features.py --field=<field_number> --whole-field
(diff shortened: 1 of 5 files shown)


mcoughlin (approved):
LGTM

PR #293, opened 2023-03-10
Enable get_features.py loop for multiple ccd/quadrants

bfhealy:
This PR enables `get_features.py` to get features for multiple ccd/quadrant combinations. Previously it could either cover all of a field's sources or only a single ccd/quad combo. This PR also simplifies the ccd/quad arguments for `get_quad_ids.py`.


mcoughlin (approved):
LGTM!

PR #296, opened 2023-03-10
Add GCE submodule

bfhealy:
This PR adds [GCE](https://github.com/mikekatz04/gce.git) as a submodule of scope. GCE cannot currently be pip-installed, and it is needed to run the CUDA-accelerated GCE algorithms in `periodsearch.py`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "scope-phenomenology"]
 	path = scope-phenomenology
 	url = https://github.com/bfhealy/scope-phenomenology.git
+[submodule "gce"]
+	path = gce
+	url = https://github.com/mikekatz04/gce.git


mcoughlin:
@bfhealy For now, we are using conditional entropy from periodfind right? Nothing wrong with pinning the repo.

bfhealy:
@mcoughlin Depending on the algorithm in `periodsearch.py`, the CE code is either from `periodfind` or `gce`. For the algorithm I think was used in the Scope I paper ('GCE_LS_AOV'), it's currently using `gce` code. I have run into some memory issues on Expanse using that algorithm though, so it might be better to use the `periodfind` CE code instead.

It's similar for LombScargle, with a mix of `cuvarbase` and `periodfind` being used depending on the name of the algorithm. 

Among the available algorithms in `periodsearch.py`, are there any you think are preferable to use for now?

mcoughlin:
@bfhealy I suggest LombScargle, CE, and AOV from periodfind. Let's see if those pass reasonably on Expanse.

bfhealy:
@mcoughlin Sounds good - those all seem to run fast on Expanse. Should I run feature generation separately for each of those 3 algorithms, or sequence them together as is currently done with the gce and cuvarbase algorithms?

mcoughlin:
@bfhealy i think it's whatever is easiest to organize data product wise. Probably cleanest is just having it just run once in a loop (and specified from the config), but it's also about what is plausible for you.

bfhealy:
@mcoughlin Ok, I'll see what I can implement to run each algorithm in the cleanest way possible. Within the feature generation script we could loop over each config-specified `periodfind` algorithm to get up to 3 periods (and associated Fourier stats for each one).

mcoughlin:
@bfhealy sounds good!

PR #297, opened 2023-03-10
WIP: add requirements for running feature generation on HPC cluster

bfhealy:
This PR adds a `requirements-hpc.txt` file that contains additional requirements to install when running scope feature generation on an HPC cluster.


mcoughlin:
@bfhealy similarly, we are using periodfind for LombScargle as well, right?

PR #302, opened 2023-03-14
Replace periodsearch error with warning when significance is nan

bfhealy:
This PR modifies the behavior of `periodsearch.py` to issue a warning instead of raising an error when a source's period significance is `nan`. The error is occasionally raised for the `EAOV` algorithm in `periodfind`. While this change is not a permanent solution, it allows the feature generation script for a field/ccd/quad to continue.

We could additionally yield a `nan` period for sources where the AOV significance is `nan` - our downstream `lcstats.py` code should be able to handle that by returning `nan` for the Fourier features dependent on that period. Currently a finite period is returned even when the significance is `nan`.

### DIFF
--- a/tools/featureGeneration/periodsearch.py
+++ b/tools/featureGeneration/periodsearch.py
@@ -1,6 +1,7 @@
 import numpy as np
 from scope import utils
 import fast_histogram
+import warnings
 
 
 def find_periods(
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #303, opened 2023-03-14
Loop over config-specified period algorithms during feature generation

bfhealy:
This PR enables feature generation to loop over period-finding algorithms specified in the config file. Depending on whether `--doCPU` or `--doGPU` is set, the code will use the corresponding algorithms in the config. Each period column in the resulting file has its algorithm appended to the column name (e.g. `period_EAOV`, `period_ECE`, etc). There are also separate Fourier stats calculated for each period.


mcoughlin (approved):
LGTM!

PR #304, opened 2023-03-15
Submit feature generation jobs in parallel

bfhealy:
This PR adds the ability for generated slurm scripts to run multiple feature generation jobs in parallel HPC instances using `sbatch`. This new functionality is in addition to the existing approach of running each job in a single interactive instance.

Also, the `period_algorithm` argument in `generate_features_slurm.py` has been removed; users can instead use the config file to specify which algorithms to run.


mcoughlin (approved):
LGTM

PR #307, opened 2023-03-15
Find periods in batches

bfhealy:
This PR modifies period finding to run in batches (default size 1000 light curves). Quadrants with many sources were raising obscure CUDA errors on Expanse when running all light curves at once, and these errors were not present for quadrants with fewer sources.


mcoughlin (approved):
LGTM!

Yay batching!

PR #315, opened 2023-03-22
Sort light curves to be monotonically increasing in time

bfhealy:
This PR sorts each light curve to be monotonically increasing in time before generating features. Multiple feature generation algorithms, including `removeHighCadence()` and `compute_dmdt()`, rely on this sorting to produce results consistent with previous generated features.


mcoughlin (approved):
Sensible. LGTM.

PR #317, opened 2023-03-24
Update feature generation documentation

bfhealy:
This PR updates documentation about `generate_features.py` and the scripts scope generates to run it.


mcoughlin (approved):
Thanks for improving the docs

PR #321, opened 2023-03-28
Draft PR for xgb.py Dreed

reed0824:
Draft PR for xgb.py Dreed,, trying to place in nn.py format.

I am missing some things here. Such as, in importing xgboost and sklearn we assume these are already installed. I can add that to the requirements.txt file, the doc.txt.  because dev gets changed depending on which mac version it is.

But, Brian, can we discuss this attempt first? Deep appreciation sent your way.

### DIFF
--- a/tools/generate_features_slurm.py
+++ b/tools/generate_features_slurm.py
@@ -280,7 +280,13 @@ if __name__ == "__main__":
         "--partition_type",
         type=str,
         default='gpu-shared',
-        help="Partition name to request",
+        help="Partition name to request for computing",
+    )
+    parser.add_argument(
+        "--submit_partition_type",
+        type=str,
+        default='shared',
+        help="Partition name to request for job submission",
     )
     parser.add_argument(
         "--nodes",
@@ -298,7 +304,13 @@ if __name__ == "__main__":
         "--memory_GB",
         type=int,
         default=180,
-        help="Memory allocation to request",
+        help="Memory allocation to request for computing",
+    )
+    parser.add_argument(
+        "--submit_memory_GB",
+        type=int,
+        default=16,
+        help="Memory allocation to request for job submission",
     )
     parser.add_argument(
         "--time",
@@ -350,6 +362,12 @@ if __name__ == "__main__":
         type=float,
         default=5.0,
         help="Time to wait between job status checks",
+    )
+    parser.add_argument(
+        "--runParallel",
+        action="store_true",
+        default=False,
+        help="If set, run jobs in parallel using slurm. Otherwise, run in series on a single instance.",
     )
 
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


bfhealy:
@reed0824 Happy to take a look! Can you please add me as a collaborator to your fork of scope? (Go to settings in your fork --> collaborators --> Add people)

scope/xgb.py
```diff
@@ -0,0 +1,206 @@
...
+from sklearn.metrics import roc_curve, auc, precision_recall_curve, classification_report, average_precision_score
+import pandas as pd
+
+import seaborn as sns
+import matplotlib.pyplot as plt
+
+import os
+import sys
+import copy
+import json
```
bfhealy, line 23: Some of these imports aren't needed although they were included in the Jupyter notebook. We can clean them up as we continue to develop the code below.

scope/xgb.py
```diff
@@ -0,0 +1,206 @@
...
+    'eta':.1,
+    'subsample': 0.7,
+    'colsample_bytree': 0.7,
+    # Other parameters
+#    'objective':'reg:squarederror',   # changed from reg:linear
+    'objective': 'binary:logistic',
+  #  'gpu_id': 0,
+  #  'tree_method': 'gpu_hist',
+    'eval_metric': 'auc'
+}
```
bfhealy, line 75: These params could be passed as individual arguments to `setup()`. For example, see the analogous code in nn.py:
```
def setup(
        self,
        dense_branch: bool = True,
        conv_branch: bool = True,
        loss: str = "binary_crossentropy",
        optimizer: str = "adam",
        callbacks: tuple = ("reduce_lr_on_plateau", "early_stopping"),
        tag: Optional[str] = None,
        logdir: str = "logs",
        **kwargs,
    ):
```
Then I'd suggest collecting the inputs in a dictionary called `params` and storing it in `self.meta`, e.g. `self.meta['params'] = params`. This will help users setup xgboost models with different parameters using this function.

scope/xgb.py
```diff
@@ -0,0 +1,206 @@
...
+    def train(
+        params,
+        dtrain,
+        num_boost_round=num_boost_round,
+        evals=[(dtest, "Test")],
+        early_stopping_rounds=10
+)
```
bfhealy, line 82: In this function we'll want to add the code in the notebook that trains the model, in the form of:
```
model = xgb.train(
    params,
    dtrain,
    num_boost_round=num_boost_round,
    evals=[(dtest, "Test")],
    early_stopping_rounds=10
)
```
(not sure whether perhaps some of the above arguments can instead be passed to `setup()`)

The first argument of any method in the XGB class should be `self`. Then we can access the info we've stored in self in previous steps, e.g. the `self.meta['params']` we set in `setup()`. 

Since scope.py already splits the learning set into train/validation/test samples, within `train()` we'll just need to convert those to the `DMatrix` objects that XGBoost uses to contain the train and test sets, e.g. from the notebook:
```
dtrain = xgb.DMatrix(X_train, label=y_train)
dtest = xgb.DMatrix(X_test, label=y_test)
```

scope/xgb.py
```diff
@@ -0,0 +1,206 @@
...
+        path = pathlib.Path(output_path)
+        if not path.exists():
+            path.mkdir(parents=True, exist_ok=True)
+
+        output_name = self.name if not tag else f"{self.name}.{tag}"
+        if not output_name.endswith('.h5'):
+            output_name += '.h5'
+        self.model.save(path / output_name, save_format=output_format)
+
+    def hist_rates(self,preds):
```
bfhealy, line 113: While the plotting will be a useful diagnostic tool going forward with this algorithm, we can first focus on getting the tools associated with training and inference running smoothly.

PR #322, opened 2023-03-29
Streamline query/analysis of nearby sources in Gaia

### LINKED ISSUE #308: Separate some Kowalski queries from main feature generation script
For fields/ccd/quads near the galactic plane with many sources, the `drop_close_bright_stars` portion of our feature generation script (querying Gaia EDR3 for sources within 300 arcsec of each ZTF source) can take several hours. This is a significant amount of time where GPUs are not needed, but are currently being used anyway. If we ran this part of feature generation separately on CPUs and saved the results, we could reduce compute cost and dedicate more of the 48-hour maximum GPU walltime to period finding.

bfhealy:
This PR reduces the amount of time the feature generation script takes to query/identify nearby sources in Gaia that are too close to ZTF sources. Only one Gaia cone search is performed across the entire quadrant (as defined by the most extreme coordinates among sources on the list + a default 300 arcsec cushion).  In addition, a ~10x performance increase is realized by using `astropy.coordinates.angular_separation()` instead of `SkyCoord.separation()` to compute angular separations. This requires that latitudes/longitudes be calculated before looping over ZTF sources.

For the most crowded field among the 20 test fields (488), the entire process now runs in ~4 min, improving from a time of multiple hours. Closes #308

### DIFF
--- a/tools/generate_features.py
+++ b/tools/generate_features.py
@@ -16,14 +16,13 @@ from scope.utils import (
 import numpy as np
 from penquins import Kowalski
 import pandas as pd
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import SkyCoord, angular_separation
 import astropy.units as u
 from datetime import datetime
 from tools.featureGeneration import lcstats, periodsearch, alertstats, external_xmatch
 import warnings
 from cesium.featurize import time_series, featurize_single_ts
 import json
-
 
 BASE_DIR = pathlib.Path(__file__).parent.parent.absolute()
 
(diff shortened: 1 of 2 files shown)


tools/generate_features.py
```diff
@@ -88,174 +86,142 @@ def drop_close_bright_stars(
...
-                    "filter": {},
-                },
-            }
+    q = kowalski_instance.query(query)
 
-            q = kowalski_instance.query(query)
-            gaia_results = q['data'][catalog]
-            gaia_results_dct.update(gaia_results)
+    gaia_results = q.get('data')
+    gaia_results_dct.update(gaia_results['Gaia_EDR3'])
```
mcoughlin, line 143: Maybe make the chosen data release a variable?

tools/generate_features.py
```diff
@@ -88,174 +86,142 @@ def drop_close_bright_stars(
...
-                xmatch_coord = SkyCoord(xmatch_ra, xmatch_dec, unit=['deg', 'deg'])
-            else:
-                xmatch_coord = SkyCoord(
-                    ra_geojson + 180.0, dec_geojson, unit=['deg', 'deg']
-                )
+    if len(query_result) > 0:
+        source_coords = np.array(
+            [x['radec_geojson']['coordinates'] for x in id_dct.values()]
+        )
+        source_coords[:, 0] += 180.0
```
mcoughlin, line 154: Should this be np.mod()?

tools/generate_features.py
```diff
@@ -88,174 +86,142 @@ def drop_close_bright_stars(
...
+            single_result = query_result.copy()
+
+            Coords = np.copy(Query_Coords)
+
+            val = id_dct[id]
+            ra_geojson, dec_geojson = val['radec_geojson']['coordinates']
+
+            # Compute separations in radians, convert to arcsec
+            # ~10x faster than SkyCoord.separation() if lon/lat is calculated out of loop
+            all_separations = angular_separation(lon1[i], lat1[i], lon2, lat2) * 206265
```
mcoughlin, line 178: Mention what 206265 is.

mcoughlin (approved):
Small comments.

tools/generate_features.py
bfhealy: Agreed!

tools/generate_features.py
bfhealy: Revised the comment above.

tools/generate_features.py
bfhealy: This line and line 162 address the [-180, 180] range of geojson RA values. I think it should stay as-is - an added constant to get to a [0, 360] range.

PR #323, opened 2023-03-29
Ensure light curves are queried in batches

bfhealy:
For multi-core feature generation, this PR decreases the size limit for light curve queries until it is small enough to run batch  queries. Otherwise, the default `query_size_limit` of 10000 sources may be greater than the number of light curves to query.

### DIFF
--- a/tools/generate_features.py
+++ b/tools/generate_features.py
@@ -348,14 +348,15 @@ def generate_features(
     print('Getting lightcurves...')
     # For small source lists, shrink LC query limit until batching occurs
     lc_limit = limit
-    while lc_limit > len(feature_gen_source_list):
-        lc_limit //= 10
+    if Ncore > 1:
+        while lc_limit > len(feature_gen_source_list):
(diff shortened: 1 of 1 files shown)


tools/generate_features.py
```diff
@@ -345,16 +343,20 @@ def generate_features(
...
-        limit=limit,
-        Ncore=Ncore,
     )
 
     print('Getting lightcurves...')
+    # For small source lists, shrink LC query limit until batching occurs
+    lc_limit = limit
+    if Ncore > 1:
+        while lc_limit > len(feature_gen_source_list):
+            lc_limit //= 10
```
mcoughlin, line 353: Shouldnt this be something like lc_limit  = np.floor(len(feature_gen_source_list) / Ncore) or something? Or is the // 10 still the best thing?

mcoughlin (commented):
just a question

mcoughlin (approved):
LGTM

PR #326, opened 2023-03-30
Refine slurm script generation

bfhealy:
This PR refines slurm script generation based on trial-and-error with the Expanse cluster.


mcoughlin (approved):
LGTM

PR #327, opened 2023-03-30
Working on setup and train functionality

reed0824:
I've updated the setup() function, and added train/call/summary/test to the class. Any comments are welcome if you have time, Brian.

I have temporarily removed plotting functionality

### DIFF
--- a/scope/xgb.py
+++ b/scope/xgb.py
@@ -1,0 +1,129 @@
+import datetime
+import os
+import pathlib
+from .models import AbstractClassifier
+
+#!git clone https://github.com/DTrimarchi10/confusion_matrix.git
+import numpy as np
+from xgboost import XGBClassifier
+import xgboost as xgb
+from sklearn.model_selection import train_test_split
+from sklearn.metrics import accuracy_score, confusion_matrix, matthews_corrcoef, mean_absolute_error
+from sklearn.metrics import f1_score, ConfusionMatrixDisplay
+#plot_confusion_matrix, is deprecated, so import ConfusionMatrixDisplay instead
+from sklearn.metrics import roc_curve, auc, precision_recall_curve, classification_report, average_precision_score
+import pandas as pd
+
+import seaborn as sns
+import matplotlib.pyplot as plt
+
+import os
+import sys
+import copy
+import json
+
+    
+
+class XGB(AbstractClassifier):
+    """Baseline model with a statically-defined architecture"""
+    def __init__(
+        self,
+        **kwargs,
+    )
+
+
+#    def call(self, inputs, **kwargs):
+#        features_input = inputs.get("features")
+#        dtrain = xgb.DMatrix(X_train, label=y_train)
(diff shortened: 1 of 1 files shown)


scope/xgb.py
```diff
@@ -0,0 +1,129 @@
...
+import json
+
+    
+
+class XGB(AbstractClassifier):
+    """Baseline model with a statically-defined architecture"""
+    def __init__(
+        self,
+        **kwargs,
+    )
```
bfhealy, line 32: Probably can remove this `__init__` method. `setup` will take care of all the initializations.

scope/xgb.py
```diff
@@ -0,0 +1,129 @@
...
+#    def call(self, inputs, **kwargs):
+#        features_input = inputs.get("features")
+#        dtrain = xgb.DMatrix(X_train, label=y_train)
+#        dtest = xgb.DMatrix(X_test, label=y_test)
+#
+#        return features_input, dtrain, dtest
+#
+    def summary(self, **kwargs):
+
+        model = xgb.train(
```
bfhealy, line 44: Remember to use `self.model` here, in addition to any other places where you'd like to set an attribute (variable accessible by all the different methods in the `XGB` class).

scope/xgb.py
```diff
@@ -0,0 +1,129 @@
...
+    def setup(
+        self,
+        'max_depth':6,
+        'min_child_weight': 1,
+        'eta':.1,
+        'subsample': 0.7,
+        'colsample_bytree': 0.7,
+        'objective': 'binary:logistic',
+        'eval_metric': 'auc',
+        **kwargs,
```
bfhealy, line 64: Nice, I think this is a good way to initialize the model parameters.

scope/xgb.py
```diff
@@ -0,0 +1,129 @@
...
+    'eval_metric': eval_metric,
+    }
+    
+    self.meta['params'] = params
+
+    self.model.compile(
+        params =self.meta["params"],
+        train = self.dtrain,
+        test = self.dtest,
+        )
```
bfhealy, line 87: I don't see `compile` in the XGBoost notebook. Should it be included here or at all?

scope/xgb.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def train(self,X_train, **kwargs):
+        dtrain = xgb.DMatrix(X_train, label=y_train)
+
```
bfhealy, line 91: This is a good place to add something like `self.model = xgb.train(...)`.

scope/xgb.py
```diff
@@ -0,0 +1,129 @@
...
+        return dtrain
+
+    def test(self, X_test, **kwargs);
```
bfhealy, line 94: I'm not sure if we need a separate test method. Perhaps it's best for now to define `dtest` like you do for `dtrain` within the `train` method, then pass it in to the call to `xgb.train(...)`, e.g.  using `evals=[(dtest, "Test")]`. You'd need to add `Xtest` as another argument to the `train` method.

PR #328, opened 2023-03-31
Remove old period algorithms

bfhealy:
This PR removes GPU period algorithms that are no longer used in scope feature generation. This code will be tested on a GPU before merging, since our GH Actions tests only run on a CPU.


mcoughlin (approved):
This should hopefully make long term maintenance easier:

PR #329, opened 2023-03-31
Clean lcstats and remove doParallel from feature generation

bfhealy:
This PR cleans code in `lcstats.py` and removes the now-unused `doParallel` from feature generation (if `Ncore > 1`, multiprocessing occurs). It also removes a duplicate multiplication of `Ncore` and `limit` in calls to the alert and external catalog queries.


mcoughlin (approved):
Nice!

PR #331, opened 2023-04-02
Add XGBoost algorithm to scope

reed0824:
In collaboration with @reed0824, this PR adds the XGBoost algorithm to scope. Once this PR is merged, the scope code will support XGB training but may need some more work to be compatible with inference. A test is added for XGB training, and four diagnostic plots are generated for each classifier:
- Confusion matrix
- Feature importance
- Precision-recall
- ROC curve


scope/xgb.py
```diff
@@ -0,0 +1,107 @@
...
+    self.meta['params'] = params
+    self.meta['evals'] = evals
+
+
+    def train(self,X_train, X_test, **kwargs):
+
+        self.model = xgb.train(self,params,evals, **kwargs)
+        
+        final_train = xgb.DMatrix(dtrain, label=y_train)
+        final_test = xgb.DMatrix(dtest, label=y_test)
```
bfhealy, line 74: We may want to use `DMatrix` before calling `xgb.train` to get the input data in the right format. I'm currently reviewing how `scope.py` manages training/validation/test datasets to integrate XGB with the current code smoothly.

scope.py
```diff
@@ -693,18 +693,31 @@ def train(
...
 
-            # X_test = ds.df_ds.loc[indexes['test']]
-            # y_test = ds.target.loc[indexes['test']]
+            X_test = ds.df_ds.loc[indexes['test']][features]
+            y_test = ds.target.loc[indexes['test']][features]
 
             # Add code to train XGB algorithm
-            XGB()
+            #XGB()
+                classifier.train(
```
bfhealy, line 710: Somewhere in our XGB-specific code, we'll need to initialize `classifier` as an `XGB` object (with a `name` = `tag`).

bfhealy:
Hi @reed0824, I made some more modifications to the code so training can run. These include initializing the `XGB` object with your `setup` method before running training, reorganizing the setup parameters to align with how XGBoost `train` expects them, and creating more distinctions between DNN and XGB in `scope.py`.

There are still some to-do items indicated by inline comments and/or `pass` in the code. Among them:
- Code a separate evaluation of the test data in the XGB `evaluate` method
- Call this evaluation in `scope.py` after training
- Remove test set evaluations from the `train` method
- Rework the XGB `save` method to save each trained model (use Ashish's notebook for guidance)

As always, let me know if you have any questions. We're getting there!

bfhealy:
@mcoughlin The test failure is related to maintenance on the latest ZTF_sources catalog on `melman`.

scope.py
```diff
@@ -529,10 +532,16 @@ def train(
...
         from scope.nn import DNN
+        from scope.xgb import XGB
         from scope.utils import Dataset
 
         train_config = self.config["training"]["classes"][tag]
 
+        if algorithm in ['DNN', 'NN', 'dnn', 'nn']:
+            algorithm = 'dnn'
+        elif algorithm in ['XGB', 'xgb', 'XGBoost', 'xgboost', 'XGBOOST']:
+            algorithm = 'xgb'
```
mcoughlin, line 543: We should raise an error if we don't find one?

scope.py
```diff
@@ -599,147 +609,207 @@ def train(
         run_eagerly = kwargs.get("run_eagerly", False)
         pre_trained_model = kwargs.get("pre_trained_model")
         save = kwargs.get("save", False)
+        plot = kwargs.get("plot", False)
         weights_only = kwargs.get("weights_only", False)
 
+        # xgb-specific arguments
```
mcoughlin, line 615: Can we add some code / document these?

scope/utils.py
```diff
@@ -764,6 +911,11 @@ def __init__(
         self.verbose = verbose
         self.target = None
 
+        if algorithm in ['DNN', 'NN', 'dnn', 'nn']:
+            self.algorithm = 'dnn'
+        elif algorithm in ['XGB', 'xgb', 'XGBoost', 'xgboost', 'XGBOOST']:
+            self.algorithm = 'xgb'
```
mcoughlin, line 917: Raise same error here?

mcoughlin (approved):
A few comments, but very solid!

scope.py
bfhealy: Agreed, will implement in the next push!

scope.py
bfhealy: I'll add some comments in the code describing these parameters in more detail.

bfhealy (approved):

PR #334, opened 2023-04-05
Generate features for fields specified in config

bfhealy:
This PR enables feature generation on user-specified lists of fields, supplementing the default behavior that randomly selects quadrants from all available fields.

A fix is included that removes the code's ability to call feature generation on the same quadrant multiple times. This is unlikely early in the process but becomes more probable as the number of remaining quadrants decreases. The code also now terminates its loop of feature generation calls before performing any duplication at the very end of the process. When the number of available instances equals the number remaining jobs to be done, the code will stop looping after it submits these last jobs. 

There remains the possibility that a dwindling HPC allocation will hinder this code, since slurm will refuse jobs that request more time than remains on the allocation (even if the jobs run in less time than requested). This code does not currently recognize these refusals and will assume that each job that it attempted to submit did get queued.

### DIFF
--- a/config.defaults.yaml
+++ b/config.defaults.yaml
@@ -1340,6 +1340,10 @@ feature_generation:
       - ECE
       - EAOV
       - ELS
+  fields_to_run:
+    # If this list is populated, generate_features_job_submission.py will only run on the specified fields.
+    - 296
+    - 297
   cesium_features:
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
Nice.

PR #337, opened 2023-04-11
WIP: add support for XGBoost inferencing

bfhealy:
This PR adds support for inference using trained XGB models. The code runs and is being tested on a ZTF field to ensure that DNN inference is unchanged while XGB inference runs correctly.

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -1238,19 +1238,19 @@ class Scope:
                     ./scope.py select_al_sample --fields=[296,297] --group='experiment' --min_class_examples=500 --select_top_n --include_all_highprob_labels --probability_threshold=0.7 --exclude_training_sources --read_consolidation_results
         """
         base_path = pathlib.Path(__file__).parent.absolute()
-        preds_path = base_path / 'preds'
-
-        # Strip extension from filename if provided
-        al_filename = al_filename.split('.')[0]
-        AL_directory_path = str(base_path / f'{al_directory}_{algorithm}' / al_filename)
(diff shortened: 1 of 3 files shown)


scope.py
```diff
@@ -1010,7 +1014,7 @@ def create_inference_script(
                 algorithm = 'xgb'
                 script.write('echo "xgb inference"\n')
                 for tag in model_tags:
-                    tag_file_gen = (group_path / tag).glob('*.h5')
+                    tag_file_gen = (group_path / tag).glob('*.json')
```
mcoughlin, line 1017: Should this glob both?

scope/xgb.py
```diff
@@ -86,7 +86,7 @@ def evaluate(self, X_test, y_test, **kwargs):
         return self.model.eval(dtest, 'dtest', **kwargs)
 
     def load(self, path_model, **kwargs):
-        self.load_model(path_model, **kwargs)
+        self.model.load_model(path_model, **kwargs)
```
mcoughlin, line 89: Any need to try except for both here?

mcoughlin (approved):
A few small questions. LGTM.

scope.py
bfhealy: So far it seems XGB models are best saved as json files, and that's the only format our code saves them in. For now I think it's appropriate to only glob json extensions for the XGB algorithm.

scope/xgb.py
bfhealy: That's a good idea, I'll incorporate a try/except loop into this method.

PR #338, opened 2023-04-12
Set ZTF DR16 as default sources catalog

bfhealy:
This PR sets the newly available `ZTF_sources_20230309` collection on `melman` as the default sources catalog in the config file. A typo in an argparse description is also fixed.


mcoughlin (approved):
LGTM

PR #342, opened 2023-04-13
Add execute permissions to scripts

bfhealy:
This PR removes the need to update script permissions each time the repo is synced by adding `+x` permissions to scripts.


mcoughlin (approved):
LGTM

PR #346, opened 2023-04-14
Add option to scale min period/max frequency for period finding

bfhealy:
This PR replaces the `--doLongPeriod` flag with `--doScaleMinPeriod` and changes its behavior, from enforcing a fixed range of periods to scaling the minimum period (and thus maximum frequency) based on the cadence set by `min_cadence_minutes`. 

If `--doScaleMinPeriod` is set and `min_cadence_minutes` = $cadence$, the maximum frequency passed to period finding (modulo units) will be $1 / (2 * cadence)$, or the Nyquist frequency for that cadence.


mcoughlin (approved):
Thank goodness someone is professionalizing this stuff ;)

PR #351, opened 2023-04-19
Add hyperparameter grid search to XGBoost training

bfhealy:
This PR adds a hyperparameter cross-validation grid search to XGBoost, with the range of each grid customizable in the config file. The optimal parameters are saved/loaded from a new .params file that accompanies the saved model. The `load_config()` method has been tested, and it successfully loads the parameters and associates them with an instantiated model. 

As in the original jupyter notebook running this code, the grid search begins with the `max_depth` and `min_child_weights` parameters and continues with `subsample` and `colsample_bytree`, then `eta`, then one more round of `max_depth` and `min_child_weights`.

### DIFF
--- a/scope.py
+++ b/scope.py
@@ -535,7 +535,7 @@ class Scope:
         from scope.xgb import XGB
         from scope.utils import Dataset
 
-        train_config = self.config["training"]["classes"][tag]
+        label_params = self.config["training"]["classes"][tag]
         train_config_xgb = self.config["training"]['xgboost']
 
         if algorithm in ['DNN', 'NN', 'dnn', 'nn']:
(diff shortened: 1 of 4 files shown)


scope.py
```diff
@@ -536,6 +536,7 @@ def train(
         from scope.utils import Dataset
 
         train_config = self.config["training"]["classes"][tag]
```
mcoughlin, line 538: Should this variable be called something else now?

mcoughlin (approved):
One small comment.

scope.py
bfhealy: That's a good idea - I'll revise with a more accurate name.

AshishMahabal (commented):
Is the step size 1 applied only to max_depth and min_child_weight?
Ideally it should be repeated for the other pair(s).

bfhealy:
Thanks @AshishMahabal, I'll add an additional grid search block for the `subsample` and `colsample_bytree` pair of parameters.

bfhealy:
Hi @AshishMahabal, do you have any other comments after the latest commits?

bfhealy:
Merging after our discussion at the ZTF ML meeting this afternoon.

PR #352, opened 2023-04-21
Run XGBoost training in parallel with slurm scripts

bfhealy:
This PR allows XGB training to be run in parallel (e.g. on SDSC Expanse) using slurm scripts, following the format of the existing code for feature generation.


mcoughlin (approved):
LGTM!

PR #355, opened 2023-04-25
Enable feature generation for user-specified sources

bfhealy:
This PR adds (and in some cases, re-implements) functionality to `generate_features.py` to generate features for ZTF light curves with IDs from a config-specified file. A new `--doSpecificIDs` flag is added to the script to trigger the workflow.

The older query-by-ID approach (removed in #322) is employed to identify sources too close to bright stars in Gaia EDR3. As a solution to avoid wasting time on HPC resources, the script can now save the output from this query/analysis and then skip that step in subsequent runs (set `--skipCloseSources`). The faster approach is still used when generating features for a given field/cdd/quadrant.

This code will be especially useful for generating up-to-date features for our training set and testing period-finding algorithms.

### DIFF
--- a/config.defaults.yaml
+++ b/config.defaults.yaml
@@ -1333,10 +1333,13 @@ feature_stats:
 
 # Below, specify algorithms, external catalogs and features to include in generated feature lists:
 feature_generation:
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #356, opened 2023-04-27
Generate diagnostic plots for DNN

bfhealy:
This PR add 3 new diagnostic plots for DNN training (CM, precision-recall and ROC) using the same code as XGBoost.


mcoughlin (approved):
LGTM

PR #357, opened 2023-04-27
WIP: Enable nested ELS_ECE_EAOV period finding

bfhealy:
This PR refactors `periodsearch.py` to enable a nested ELS - ECE - EAOV approach on GPUs.

### DIFF
--- a/config.defaults.yaml
+++ b/config.defaults.yaml
@@ -1344,6 +1344,7 @@ feature_generation:
     CPU:
       - LS
     GPU:
+      # - ELS_ECE_EAOV
       - ECE
       - EAOV
       - ELS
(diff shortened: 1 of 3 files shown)


mcoughlin:
@bfhealy do you feel like we need it? Aren't we running all algorithms at full frequency grid now?

bfhealy:
@mcoughlin I'm not completely sure -  we are indeed running each algorithm on the full frequency grid, but Matthew suggested we might improve our performance by nesting the algorithms. I thought I'd experiment using this PR and Expanse.

I think the primary means of improvement will be to expand the `freqs_to_remove` list in the script, e.g. with multiples of 1 year. I've found that for the known periodic sources, ~2/3 of ECE and EAOV periods from DR16 are within 10% of the period values in the golden set. Significantly fewer of the ELS periods (~7%) are within 10% of the golden set periods. I'll make sure to post an issue about this too.

mcoughlin:
@bfhealy Ok as long as both separate and nested remain possible

bfhealy:
Closing in favor of #358.

PR #362, opened 2023-05-03
Add nested ELS_ECE_EAOV period algorithm with minimal extra compute

bfhealy:
This PR adds a GPU-accelerated nested period-finding algorithm that determines the best EAOV significance for the top N periods from the ELS and ECE algorithms. It avoids the need to run EAOV a second time by accessing a subset of the results from the full EAOV run that was already performed.

Specifying this nested approach in the config file will result in four periods/significances/pdots being saved: those corresponding to the original ELS, ECE and EAOV algorithms along with new ELS_ECE_EAOV values.

Note: the scope test suite will fail while `melman` is unavailable.


config.defaults.yaml
```diff
@@ -1344,9 +1344,12 @@ feature_generation:
     CPU:
       - LS
     GPU:
-      - ECE
-      - EAOV
-      - ELS
+      # If ELS_ECE_EAOV is set, period finding will be run using a nested approach which passes the top N ELS and ECE periods to EAOV.
+      # Individual ELS, ECE and EAOV periods will also be saved.
+      - ELS_ECE_EAOV
```
mcoughlin, line 1349: Maybe this is better as an option rather than a default until some study is done? Or are you confident this has improved performance?

tools/featureGeneration/periodsearch.py
```diff
@@ -226,9 +223,9 @@ def find_periods(
                     period = periods[np.argmin(stat.data)]
                 elif algorithm.split("_")[0] == "EAOV":
                     significance = np.abs(
-                        np.mean(stat.data) - np.min(stat.data)
+                        np.mean(stat.data) - np.max(stat.data)
```
mcoughlin, line 226: Nice catch.

mcoughlin (approved):
Just a question on making this the default or not.

config.defaults.yaml
bfhealy: That might be a good idea, especially since my preliminary check showed minimal differences between the golden set periods and the new algorithm's performance as compared to the full EAOV run. 92% of the ELS_ECE_EAOV periods are within 10% of the corresponding period from the full EAOV run.

tools/featureGeneration/periodsearch.py
bfhealy: Thanks!

config.defaults.yaml
mcoughlin: Ok. If you look at them and decide those 10% are mostly better with this new method, then you might go for it.

bfhealy: Sounds good!

mcoughlin:
@bfhealy YEah we can fix that later.

PR #363, opened 2023-05-03
Revise docs, delete another .DS_Store

bfhealy:
This PR revises our documentation with recent updates and deletes another .DS_Store file that made its way into the repo a while ago.


mcoughlin (approved):
LGTM

PR #364, opened 2023-05-04
Add 0.2d to period-finding mask

bfhealy:
This PR adds a frequency range corresponding to 0.2 d to our `--doRemoveTerrestrial` period-finding mask. After reading https://arxiv.org/abs/2304.13843, I found that we already mask most of the suggested period ranges (with some additional padding), but we were missing an 0.2d mask.


mcoughlin (approved):
LGTM

PR #365, opened 2023-05-04
Update slurm script generation with new arguments

bfhealy:
This PR updates `generate_features_slurm.py` to handle arguments recently added to the feature generation script and the job submission code.


mcoughlin (approved):
LGTM

PR #366, opened 2023-05-05
Pin periodfind 0.0.5

bfhealy:
This PR pins the latest version of periodfind.


bfhealy:
Agreed about the test. I can work on a fix by commenting/modifying the hardcoded references to melman that currently fail. Not sure if we're close to a new penquins release - if so, it would be a simpler fix.

PR #330, opened 2023-03-31
Update kowalski queries to use new penquins multi-instance querying

bfhealy:
This PR updates several scripts in scope to simplify interactions with Kowalski (specifically, the `kowalski`, `gloria` and `melman` instances). The code provides compatibility with the latest merged `penquins` changes (https://github.com/dmitryduev/penquins/pull/23). The config is updated to require a token for each listed Kowalski instance. After that, there is no longer a need to specify which instance to query in the rest of the scope code.

Additionally, a new keyword in `get_lightcurves_via_ids` allows a smaller subset of data to be returned from light curve queries - only what's needed for feature generation.  

This PR will pass tests once a new release of `penquins` becomes available.

PR #368, opened 2023-05-05
Fix failing test suite by removing melman references

bfhealy:
This PR fixes the failing scope test suite by switching hardcoded `melman` references to `gloria`. This will need to be undone once `melman` is available in order for feature generation to run on the correct catalog. Once the latest version of `penquins` is released, the code will no longer need these references to specific instances (see #330).


mcoughlin (approved):
LGTM

PR #369, opened 2023-05-05
Add additional hyphen to arguments in download/upload/annot scripts

bfhealy:
Our Fritz download/upload/annotation scripts currently take command line arguments with one hyphen, e.g. `-file`. For the sake of consistency with our other scripts that take arguments (as well as Unix standards), this PR adds a second hyphen to all arguments longer than one character in these scripts. The documentation is updated accordingly.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -258,7 +258,7 @@ process:
 output: data with new columns appended.
 
 ```sh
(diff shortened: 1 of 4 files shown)


mcoughlin (approved):
Nice

PR #370, opened 2023-05-08
Add ability to download active learning classifications, merge with training set

bfhealy:
This PR modifies `scope_download_classification.py` to enable the expansion of the current training set by downloading active learning classifications. If `--update_training_set` is set, the code will check whether a classification's 'net votes' (the difference between number of upvotes and downvotes) is ≥ the user-specified threshold (default 1). Each classification satisfying this criterion will be included in the file produced by running the script.

Although the active learning sample is intended to be selected from sources not already in the training set, the code searches for any overlap using ZTF light curve IDs. Light curves duplicated between the training and AL samples receive updated classifications from the latter, and the original entry in the file is removed. The code places the new rows for these sources at the end of the output file with a prompt to run `scope_upload_classification.py` while setting `--start` to the appropriate index to both update existing sources and post new ones.

### DIFF
--- a/tools/scope_download_classification.py
+++ b/tools/scope_download_classification.py
@@ -16,12 +16,15 @@ from scope.utils import impute_features
 
 NUM_PER_PAGE = 500
 CHECKPOINT_NUM = 500
+BASE_DIR = pathlib.Path(__file__).parent.parent.absolute()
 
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #371, opened 2023-05-09
Update usage.md to be clearer about named arguments

bfhealy:
This PR updates the docs to explicitly list each named argument in the download, upload and manage_annotation scripts. It also adds the newest named arguments to the download documentation.


mcoughlin (approved):
LGTM

PR #373, opened 2023-05-09
Rename relphi5 to relphi4 in calc_stats and calc_fourier_stats

bfhealy:
This PR renames `relphi5` to `relphi4` in two `lcstats` functions to reduce confusion.


mcoughlin (approved):
LGTM

PR #374, opened 2023-05-10
Add XGB taxonomy mapper, modify DNN mapper

bfhealy:
This PR adds a Fritz taxonomy mapper file for labels from XGB and updates the DNN mapper to cover all current binary classifiers.


mcoughlin (approved):
LGTM

PR #375, opened 2023-05-12
Add source count to check_quads_for_sources

bfhealy:
This PR adds code to `check_quads_for_sources` that returns source counts in each ccd/quad for specified fields. This will help gauge the completeness of the incoming DR16 catalog on melman compared to DR5 on gloria.


mcoughlin (approved):
LGTM

PR #376, opened 2023-05-12
Allow catalog completeness results to be saved

bfhealy:
This PR allows a json file containing Kowalski catalog completeness results to be saved.


mcoughlin (approved):
LGTM

PR #377, opened 2023-05-15
Switch find query to count_documents in check_quads_for_sources()

bfhealy:
This PR switches a `find` Kowalski query in `check_quads_for_sources` to `count_documents`, which offers ~10x time savings over querying individual IDs.


mcoughlin (approved):
LGTM

PR #378, opened 2023-05-15
Move external catalog feature projections outside loop

bfhealy:
This PR attempts to fix a bug where generated feature files contain unneeded coordinate columns associated with external catalogs (e.g. `PS1_DR1__coordinates.radec_geojson`). This happens when generating features on enough sources to require at least two iterations over the loop in `external_xmatch.py`. The fix moves the code associated with defining the desired external features outside the loop. It will be merged after I confirm it works as intended on the full list of training sources.


mcoughlin (approved):
LGTM

PR #380, opened 2023-05-16
Fix json.dump bug in catalog completeness code

bfhealy:
This PR converts dictionary keys in `check_quads_for_sources` to `str` to avoid a `json.dump` error when saving.


mcoughlin (approved):
LGTM

PR #381, opened 2023-05-16
Add minobs filter to catalog completeness code

bfhealy:
This PR adds a `minobs` argument to `check_quads_for_sources`. Setting this argument to the minimum number of light curve points enforced in scope (currently 50) will help provide more accurate feature generation runtime estimates. This is because the longest-running part of feature generation, the period finding, only runs on sources having ≥ `minobs` points. 

Also in this PR, a `$gt` is changed to `$gte$` in `get_quad_ids.py` to be consistent with the stated strategy of using light curves having N or more points.


mcoughlin (approved):
LGTM

PR #384, opened 2023-05-22
Enable inferencing on generated features

bfhealy:
This PR enables `inference.py` to run on feature files generated using `generate_features.py`. While this approach will not be necessary in the longer term once these features are ingested by Kowalski, it is useful in the short term. In addition, the inference script is refactored to run more efficiently by looping over all labels for a single parquet file, rather than requiring multiple function calls for each label. This minimizes the number of times features must be imputed during inference.

This PR also addresses a significant change in feature generation since it was adapted from `ztfperiodic`: the inclusion of multiple periods from different algorithms (e.g. `period_ELS`, `period_ECE`, etc.) and their associated features. Previously the code required the manual modification of each feature in the `features:` section of the config file to align with the feature names for the desired algorithm. The changes here allow user can specify a `period_suffix` during training and inference that is appended to all features marked as `periodic: true` in the config file.

The `create_training_script` and `create_inference_script` methods in `scope.py` are modified accordingly. Also, `scope.py test` is expanded to run inference on generated features (it previously only ran on features queried from Kowalski).


mcoughlin (approved):
LGTM

PR #389, opened 2023-05-24
Add field/ccd/quad/filter info to inference results

bfhealy:
This PR adds ZTF field/ccd/quad/filter fields to each row in the file produced by `inference.py`. Previous classification catalogs have lacked this information, but it could be useful when querying new classification catalogs in the future.


mcoughlin (approved):
Good idea!

PR #390, opened 2023-05-24
Save inference results as parquet instead of hdf5

bfhealy:
This PR changes the default file format for inference results to parquet. This change, combined with the explicit conversion of integer fields to pandas `Int64` type, ensures that Kowalski ingests these fields as integers rather than floats.


mcoughlin (approved):
LGTM

mcoughlin (approved):
LGTM

PR #391, opened 2023-05-24
Rename generated features index to '_id', reset index

bfhealy:
Currently in `generate_features.py`, the ZTF ID for each source is stored as the index column. This PR renames that column to `_id` (to interface smoothly with kowalski/mongoDB) and then resets the index so the `_id` column can be accessed during ingestion.


mcoughlin (approved):
LGTM

PR #393, opened 2023-05-25
Add script to download GCN sources

bfhealy:
This PR add a script (`scope_download_gcn_sources.py`) to download sources associated with a GCN event on Fritz. The script saves a json file that can be used for feature generation, facilitating a quick turnaround to running inference and posting relevant classifications. `inference.py` is also modified to accommodate this variation on existing scope workflows, and the documentation is updated.


mcoughlin (approved):
This is gonna be good.

PR #395, opened 2023-05-26
Gcn workflow updates

bfhealy:
This PR makes updates to feature generation and GCN source querying to make the workflow run more smoothly. Primarily:
- Bugfix: the page-by-page query was missing `localizationDateobs`, `startDate` and `endDate`.
- A parquet file is now generated by `scope_download_gcn_sources.py` to be used as the primary input to the feature generation script. `generate_features.py` handles this file smoothly and passes on the identifiers from Fritz if they are available.


mcoughlin (approved):
LGTM

PR #397, opened 2023-05-30
Address remaining job submission bug

bfhealy:
#392 did not adequately fix the slurm job submission bug. This PR solves the issue by restoring some of the script's original functionality but introducing a `new_max_instances` variable that defines how many instances can be run at once based on how many jobs remain. This change keeps things running smoothly regardless of whether more or fewer jobs remain than `args.max_instances`.


mcoughlin (approved):
LGTM

PR #405, opened 2023-06-02
Update sweep defaults, round DNN inference probabilities

bfhealy:
This PR updates wandb sweep defaults and rounds DNN inference probabilities to 2 decimal places (replicating previous behavior).


mcoughlin (approved):
LGTM

PR #406, opened 2023-06-02
Update slurm script generation and training workflow

bfhealy:
This PR generalizes the training slurm script generation and workflow (it was previously XGBoost-specific). The code now supports GPU training and running a round of sweeps before training.


mcoughlin (approved):
LGTM

PR #407, opened 2023-06-05
Add interval between job submission to troubleshoot wandb issue

bfhealy:
This PR adds an interval (default 5 s) between running training/sweep jobs in an attempt to troubleshoot a sporadic WandB issue where the run fails with the message below:
```
400 response executing GraphQL.
{"errors":[{"message":"Sweep user not valid","path":["upsertSweep"]}],"data":{"upsertSweep":null}}
wandb: ERROR Error while calling W&B API: Sweep user not valid (<Response [400]>)
wandb: ERROR Sweep user not valid
```

### DIFF
--- a/tools/train_algorithm_job_submission.py
+++ b/tools/train_algorithm_job_submission.py
@@ -52,7 +52,13 @@ def parse_commandline():
         "--wait_time_minutes",
         type=float,
         default=5.0,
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #409, opened 2023-06-07
Distinguish completed and running slurm training jobs

bfhealy:
This PR modifies `train_algorithm_job_submission.py` to save a dummy file that indicates whether a training job has been queued to run. The code now makes an important distinction between completed and running jobs. This ensures that the same classifier is not trained a second time while the first iteration is running but not yet completed. A new keyword `--reset_running` allows new runs of jobs that may have failed the first time around.

The default hyperparamers in the config are also updated to include the parameters from the WandB sweeps.


bfhealy:
On hold while updating scope for new penquins release.

mcoughlin:
@bfhealy I posed 2.3.1 for penquins last night.

mcoughlin (approved):
LGTM

PR #414, opened 2023-06-09
Adapt new job submission approach for feature generation

bfhealy:
This PR builds on #409, adapting the approach of distinguishing between running and completed jobs in `generate_features_job_submission.py`. 

Unneeded code is also removed from `train_algortihm_job_submission.py`, and a keyword to specify the model's group name (independent of the training script) is added.


mcoughlin (approved):
LGTM

PR #419, opened 2023-06-13
Add inference batch_size kwarg, address self-impute failures

bfhealy:
This PR modifies `inference.py` by adding a `batch_size` kwarg to allow the user to control the size of feature batches read during inference. It also addresses when self-imputation of features fails for a small set of sources with an all-NaN column. In these cases, the code will use the training set specified in the config file to impute missing features.


mcoughlin (approved):
LGTM

PR #420, opened 2023-06-14
Define amount of jobs to queue based on nchoice

bfhealy:
In `generate_features_job_submission.py`, this PR redefines the number of jobs to queue based on `nchoice` instead of `njobs`. This avoids an error in the submission script once all jobs have been queued.


mcoughlin (approved):
LGTM

PR #421, opened 2023-06-14
Add support for inference slurm jobs

bfhealy:
This PR adds code to run inference jobs on Expanse following the pattern of feature generation and training. A new section of the config file specifies the fields on which to run inference, and the other details are specified via `run_inference_slurm.py`. After running `slurm_submission.sub`, `run_inference_job_submission.py` checks for parquet files in each field of the predictions directory, queuing inference jobs for fields lacking results.


mcoughlin (approved):
LGTM

PR #423, opened 2023-06-16
Fix get_lightcurves_via_coords for new penquins, add to tests

bfhealy:
This PR fixes the query in `get_lightcurves_via_coords` to work with penquins 2.3.1. The PR also sets the default source catalog using the config file rather than selecting the last catalog name in a list. A call to `get_lightcurves_via_coords` is added to the tests.

### DIFF
--- a/scope/fritz.py
+++ b/scope/fritz.py
@@ -71,6 +71,7 @@ BASE_URL = f"{config['fritz']['protocol']}://{config['fritz']['host']}/"
 MAX_ATTEMPTS = config['fritz']['max_attempts']
 SLEEP_TIME = config['fritz']['sleep_time']
 fritz_token = config['fritz']['token']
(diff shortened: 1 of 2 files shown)


scope/fritz.py
```diff
@@ -71,6 +71,7 @@ def get_stats(G, ids):
 MAX_ATTEMPTS = config['fritz']['max_attempts']
 SLEEP_TIME = config['fritz']['sleep_time']
 fritz_token = config['fritz']['token']
+default_catalog = config['kowalski']['collections']['sources']
```
mcoughlin, line 74: Should this be a get in case it is not defined?

mcoughlin (approved):
LGTM. One small comment.

scope/fritz.py
bfhealy: Good idea, I'll make that change.

PR #424, opened 2023-06-16
Fix completed fields bug in inference job submission script

bfhealy:
This PR fixes code in `run_inference_job_submission.py` that modified the same list it iterated over, producing an incorrect list of completed fields.


mcoughlin (approved):
LGTM

PR #426, opened 2023-06-21
Overhaul combine_preds to facilitate Kowalski ingestion

bfhealy:
Changes to `inference.py` have rendered `combine_preds.py` obsolete in its current form. This PR overhauls `combine_preds.py` to serve the new purpose of combining DNN and XGB prediction columns into one parquet file per field. This will facilitate a single-step ingestion of DNN/XGB predictions to Kowalski, which is important since Kowalski code does not currently support update existing entries with new fields.

### DIFF
--- a/combine_preds.py
+++ b/combine_preds.py
@@ -1,43 +1,69 @@
-import fire
+#!/usr/bin/env python
 import pandas as pd
-import glob
 import os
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM!

PR #427, opened 2023-06-23
Fix feature generation query error when coordinates are near 0 RA

bfhealy:
This PR fixes an error that arose when querying Gaia sources for a ZTF field/ccd/quad with RA values near 0 degrees. The previous code did not account for the circular nature of the RA coordinate, so the difference between RAs of 0.1 deg and 359.9 deg was very large. The code now uses scipy's `circmean` to compute the circular mean of the RA coordinate. Downstream code now uses `astropy.coordinates.SkyCoord` to determine the max separation between the central ra/dec and the source coordinates.

### DIFF
--- a/tools/generate_features.py
+++ b/tools/generate_features.py
@@ -27,6 +27,7 @@ import warnings
 from cesium.featurize import time_series, featurize_single_ts
 import json
 from joblib import Parallel, delayed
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
tricky.

PR #429, opened 2023-06-27
Allow feature generation sources filepath to be specified in calls to script

bfhealy:
This PR allows the filepath for a source list to be specified in calls to `generate_features.py`, rather than only in the config file. This will be useful for running cron jobs on transient candidates.

### DIFF
--- a/tools/generate_features.py
+++ b/tools/generate_features.py
@@ -455,6 +455,7 @@ def generate_features(
     skipCloseSources: bool = False,
     top_n_periods: int = 50,
     max_freq: float = 48.0,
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

bfhealy:
@mcoughlin I think gloria's having some connection issues, so I expect the remaining tests to fail.

mcoughlin:
@bfhealy you might figure out how to separate tests that rely on external connections like that to their own run.

bfhealy:
@mcoughlin That's a good idea, I'll look into that.

PR #431, opened 2023-06-28
Fix missing algorithm kwarg in slurm training script

bfhealy:
This PR makes a copy of training script line info so the code detects the algorithm kwarg. Previously it was removed owing to list mutability, so the algorithm would default to 'dnn' even when 'xgb' was specified.


mcoughlin (approved):
LGTM

PR #432, opened 2023-06-30
Add script to automatically classify sources near GCN candidates

bfhealy:
This PR adds a script (`gcn_cronjob.py`) that automatically classifies light curves within 0.5 arcsec (by default) of candidates for GCN events. The script queries all GCN events within the last 14 days, identifies any new sources, runs feature generation on Expanse, pulls those results and runs inference locally, consolidates predictions, and uploads DNN/XGB classifications above a mean probability threshold of 0.7. A metadata file for each event lists sources and their classifications passing the threshold.

This script can be run as a cron job, e.g. on the hour:
```
0 * * * * ~/miniforge3/envs/scope-env/bin/python ~/scope/gcn_cronjob.py > ~/scope/log_gcn_cronjob.txt 2>&1
```

A future PR will allow light curve plots associated with the classifications to be posted as comments.

### DIFF
--- a/gcn_cronjob.py
+++ b/gcn_cronjob.py
@@ -1,0 +1,368 @@
+#!/usr/bin/env python
+from scope.fritz import api
+from datetime import datetime, timedelta
+import argparse
+import pathlib
+import yaml
+from tools.scope_download_gcn_sources import download_gcn_sources
+import os
+from scope.utils import read_parquet
+import numpy as np
+import warnings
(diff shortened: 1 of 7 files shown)


mcoughlin (approved):
I suspect this is a PR just best tested in the wild. Good luck ;)

bfhealy:
Thanks, I agree!

PR #439, opened 2023-07-06
Fix cronjob issues specific to running in cron environment

bfhealy:
The cron environment does not have access to the same paths and environment variables as a login shell. This PR updates `gcn_cronjob.py` to run successfully given these differences. Most notably, the path to the scope environment's python installation (with all required packages installed) can now be specified. Some background info is added as comments here, with more documentation planned for a future PR.


mcoughlin (approved):
LGTM

PR #440, opened 2023-07-07
Save training stats in json file

bfhealy:
Training stats are currently included in output plots, but there is no text file to read for easy access. This PR saves a json file containing a model's accuracy, precision, recall, f1 score, and area under ROC curve for each of the train, val and test sets.


mcoughlin (approved):
LGTM

bfhealy:
@Leighannaz Once this code is merged, I'll run another round of DNN/XGB training and upload the results with the new stats files.

PR #442, opened 2023-07-12
Add test workflow without Kowalski connection

bfhealy:
This PR adds a test workflow that does not require a Kowalski connection. It also removes balance hyperparameter values from the config defaults, since we currently train all classifiers on the full training set.


mcoughlin (approved):
LGTM

PR #446, opened 2023-07-20
GCN cronjob: save list of classified sources to reduce duplicate inference

bfhealy:
When running `gcn_cronjob.py`, the same candidates can be associated with multiple GCN events. Rather than repeatedly running feature generation/inference on sources near these candidates, this PR maintains a "checkpoint list" of classified sources and excludes them from subsequent inference unless `ignore_checkpoint` is specified. The checkpoint list keeps track of GW and GRB-associated sources separately. The list file is deleted and remade every 30 days (by default) to avoid an endlessly growing list of ids.

### DIFF
--- a/gcn_cronjob.py
+++ b/gcn_cronjob.py
@@ -11,6 +11,7 @@ import os
 from scope.utils import read_parquet
 import numpy as np
 import warnings
+import json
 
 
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Definitely a good idea.

PR #447, opened 2023-07-21
Change how inference.py determines field type

bfhealy:
To facilitate running `inference.py` in a Jupiter notebook for the ZTF summer school, this PR changes how the code determines the `type` of the `field` variable. This is necessary since the `os.system` commands in the notebook always provide the field as a string.


mcoughlin (approved):
LGTM

PR #448, opened 2023-07-21
Allow config-specified dataset path during training

bfhealy:
This PR allows the training dataset's path to be taken from `config.yaml` when running `scope.py train`. While this already happens when creating a training script, a single call to `scope.py train` (e.g. in the summer school notebook) currently requires the path to be explicitly specified.


mcoughlin (approved):
LGTM

PR #449, opened 2023-07-21
Use self.base_path in scope.py

bfhealy:
This PR uses the new `self.base_path` attribute in `scope.py` in more places where pathlib.Path(...) was previously used.


mcoughlin (approved):
LGTM

PR #453, opened 2023-08-01
Add scanner guide to docs index

bfhealy:
This PR adds the new scanner guide to the documentation index so it can be reached more easily.


mcoughlin (approved):
LGTM

PR #455, opened 2023-08-02
Add cron job info in docs

### LINKED ISSUE #428: Add documentation for running automated analyses
#386 calls for code to enable automated analysis of transient candidates using SCoPe trained models. When this work in progress is implemented, we should supplement it with details in the documentation about how this is done, both specifically and generally.

bfhealy:
This PR adds information about running SCoPe cron jobs to the usage docs. Resolves #428.


mcoughlin (approved):
LGTM

PR #456, opened 2023-08-02
Post ZTF photometry plots as comments when uploading classifications

bfhealy:
This PR adds the ability for `scope_upload_classification.py` to generate ZTF photometry plots and post these plots as comments/attachments on Fritz. It is important not to post this photometry to the source itself when that source is a transient candidate, or the time of first detection will be changed. 

Example photometry plot:
ZTF20abaousf

Example comment text and attachment on SkyPortal:
Screen Shot 2023-08-03 at 12 02 14 PM


mcoughlin (approved):
LGTM

PR #457, opened 2023-08-03
Add option to post phase-folded photometry to Fritz comments

bfhealy:
This PR adds a flag to post phase-folded photometry as an additional Fritz comment if already posting the time series.


mcoughlin (approved):
LGTM

PR #461, opened 2023-08-08
Prevent --ignore_checkpoint from duplicating checkpoint entries in gcn_cronjob.py

bfhealy:
This PR prevents the `--ignore_checkpoint` flag in `gcn_cronjob.py` from adding duplicate entries of ZTF candidate ids to the checkpoint list.


mcoughlin (approved):
LGTM

PR #464, opened 2023-08-11
Add light curve plot description to Fritz scanner guide

bfhealy:
This PR adds a line to the scanner documentation describing the light curve plots that are posted to Fritz when `gcn_cronjob.py` runs.


mcoughlin (approved):
LGTM

PR #468, opened 2023-08-15
Increase cronjob checkpoint refresh to 180 days

bfhealy:
This PR increases the default time interval to refresh the GCN cronjob checkpoint list from 30 to 180 days.


mcoughlin (approved):
LGTM

PR #469, opened 2023-08-16
Add selected feature definition tables to usage.md

bfhealy:
This PR adds tables to the Usage docs defining the phenomenological and ontological features Scope classifiers currently use.


mcoughlin (approved):
Great addition.

PR #470, opened 2023-08-17
Add method to find and copy classifier stats to new directory

bfhealy:
This PR adds a method to `scope.py` that finds all classifier performance stats and copies them to a new directory so they are more easily accessible.


mcoughlin (approved):
LGTM

PR #471, opened 2023-08-18
Update classification lists in scanner guide

bfhealy:
This PR updates the classification lists in the scanner guide to align with the current set of classifiers achieving nonzero recall and a defined level of precision.


mcoughlin (approved):
LGTM

PR #472, opened 2023-08-18
Update plotting notebook to use assemble_training_stats output

bfhealy:
This PR updates the `SCoPe_data_analysis_plots` notebook to use the collection of classifier stats output by `scope.py assemble_training_stats`. It also slightly modifies the precision/recall plots for the SCoPe paper.

Also, the `RRd` label is returned to the list in the scanner docs, since one of the two ML algorithms achieves good results for it.


mcoughlin (approved):
LGTM

PR #473, opened 2023-08-21
Plot DNN architecture during training

bfhealy:
This PR adds code to save a plot of the DNN architecture using `tf.keras.utils.plot_model`. It also adds a `pydot` requirement needed for this code to work.


mcoughlin (approved):
LGTM

PR #475, opened 2023-08-23
Update tests to work with DR16 features, classifications

bfhealy:
This PR updates the tests and config defaults to work with the DR16 feature and classification collections on `melman`. It makes some tweaks to the utils and inference code to facilitate this.


mcoughlin (approved):
LGTM

PR #477, opened 2023-08-25
Update kowalski query notebook for DR16

bfhealy:
This PR updates the notebook containing kowalski query examples to connect with melman (in additional to kowalski and gloria) and to use ZTF DR16 collections.


mcoughlin (approved):
LGTM

PR #480, opened 2023-08-31
Only apply balancing to training set, leaving val and test sets unchanged

bfhealy:
This PR changes the behavior of the `balance` hyperparameter so that only the training set is modified. The validation and test sets are left unchanged regardless of the desired balance to ensure that stats like precision and recall can be compared to each other across all training runs. 

Additionally, the `weight_per_class` parameter can now be used during XGB training by setting the `scale_pos_weight` parameter under the hood.


mcoughlin (approved):
LGTM

PR #482, opened 2023-09-01
Add new plots to data analysis notebook

bfhealy:
This PR adds new plots of XGB feature importance and DNN vs. XGB precision/recall to the data analysis notebook. It also updates the headers in the kowalski query notebook to reflect melman's inclusion among the instances.


mcoughlin (approved):
LGTM

PR #484, opened 2023-09-07
Add new analysis plots, restructure analysis notebook

bfhealy:
This PR adds new plots displaying feature importance stats, test/train score differences, and number of positive training examples to the analysis notebook. It also restructures how existing plots are generated without changing their content.


mcoughlin (approved):
LGTM

PR #487, opened 2023-09-12
Reduce annotation/instrument api calls when posting sources

bfhealy:
This PR reduces api calls to Fritz when posting new sources. Specifically, the code checks for exiting period annotations before sending a POST request, and it queries current instruments once rather than on a per-source basis.

Also, two rarely-used functions are moved to the bottom of the file.


mcoughlin (approved):
LGTM

PR #489, opened 2023-09-14
Move Fritz instrument_id request to upload script

bfhealy:
This PR moves the instrument API request from `fritz.py` to `scope_upload_classification.py`. This avoids the automatic running of this request whenever code from `fritz.py` is imported, while allowing a new `--instrument_name` argument to be passed to `scope_upload_classification.py`.


mcoughlin (approved):
LGTM

PR #491, opened 2023-09-22
Fix feature generation KeyError bug

bfhealy:
This PR fixes a bug in feature generation in which a KeyError could be raised when calculating basic stats. The bug occurred because sources that generated a separate error during initial/cesium feature generation were not always removed from all relevant dictionaries.


mcoughlin (approved):
LGTM

PR #493, opened 2023-09-27
Allow more path customization for feature generation/inference

bfhealy:
SCoPe code typically references a single `BASE_DIR` that contains the code along with input and output data. This PR adds more flexibility by allowing users to specify a `path_to_features` in the config file other than the base SCoPe directory. If this path is given, the `BASE_DIR` variable will be redefined in feature generation code and results will be saved to the specified directory. This is useful for generating features on a cluster, where the code's installation directory may not be ideal for saving larger files. 

Additionally, this PR makes use of the existing `path_to_preds` config entry to allow inference results to be saved in a directory other than the base code directory. Further testing will be done on Expanse to ensure the path updates work as intended.

(See also #488)

### DIFF
--- a/config.defaults.yaml
+++ b/config.defaults.yaml
@@ -1358,6 +1358,8 @@ feature_stats:
 
 # Below, specify algorithms, external catalogs and features to include in generated feature lists:
 feature_generation:
+  # Path to save generated features
+  path_to_features:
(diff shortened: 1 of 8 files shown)


mcoughlin (approved):
LGTM

PR #498, opened 2023-10-10
Standardize installation across platforms

### LINKED ISSUE #347: Python 3.11 install issues
We should mention in the setup guide that some packages are not yet Python 3.11 compatible. We should instead recommend installing with 3.10.

### LINKED ISSUE #494: Adjust docs to reflect easier tensorflow installation
Tensorflow 2.13 is easier than previous versions, even on arm64 Macs (`pip install tensorflow`). We should update the documentation and installation process to reflect this.

bfhealy:
This PR standardizes the scope installation process across all platforms, removing `dev-M1.txt` and instead installing tensorflow with pip. The documentation is updated to reflect the new installation steps, and tests are now run using Python 3.11. 

(Resolves #347, resolves #494)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,7 +26,7 @@ jobs:
 
     - uses: actions/setup-python@v4
       with:
-        python-version: '3.9'
+        python-version: '3.11'
 
     - name: Install dependencies and use default config
       run: |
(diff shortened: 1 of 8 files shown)


mcoughlin (approved):
LGTM

PR #503, opened 2023-10-31
Fix CPU AOV algorithm bug, parallelize CE

bfhealy:
This PR fixes a bug in the CPU AOV period-finding code and parallelizes the CPU CE algorithm, which currently runs the slowest. The CE algorithm is now tested along with the LS algorithm in `config.defaults.yaml`. AOV is not tested because of the additional compilation/installation required.


mcoughlin (approved):
LGTM

PR #508, opened 2023-11-08
Add max_timestamp_hjd as argument in generate_features.py

bfhealy:
This PR allows `--max_timestamp_hjd` to be specified as an argument in `generate_features.py`, bringing it in line with other arguments that can be specified both in `config.yaml` and the feature generation script.


mcoughlin (approved):
LGTM

PR #511, opened 2023-11-21
WIP: Enable local feature generation/inference in single script

bfhealy:
This PR adds a script that runs feature generation/inference locally based on an input file of ra/dec and/or ZTF light curve IDs. WIP while adding code to run inference inline rather than through a shell script.


mcoughlin (approved):
Sorry @bfhealy. LGTM!

bfhealy:
No worries, thanks!

PR #513, opened 2023-12-08
Allow combine_preds.py to access config-specified fields

bfhealy:
This PR allows `tools/combine_preds.py` to access the inference fields specified at the end of `config.yaml` by adding the `--use_config_fields` argument. Since we tend to run inference and then combine DNN/XGB predictions for the same subsets of fields at a time, this will make it easier to expand scope classification results without duplicating compute effort on completed fields.


mcoughlin (approved):
LGTM

PR #516, opened 2023-12-12
add plots showing scores vs positive examples

jillhughes141:
This pr adds new plots to the Jupyter notebook showing the relationship between precision and recall scores vs number/positive training examples.


bfhealy (approved):
Thanks Jill, these are helpful new plots to have in the notebook!

PR #521, opened 2023-12-15
Check for completeness file before loading

bfhealy:
This PR ensures that a completeness file exists before attempting to load it, avoiding a potential error when generating a new file.


mcoughlin (approved):
LGTM

PR #525, opened 2023-12-27
Add runtime to feature generation logs

bfhealy:
This PR adds the total runtime to the outputs logged during feature generation. This will simplify future efforts to compare runtimes and the number of initial sources per quadrant (eliminating the present need to parse slurm email subject lines for runtimes).

PR #526, opened 2024-01-03
Refactor analyze_logs to use new runtime line

bfhealy:
This PR refactors `tools/analyze_logs.py` to use the feature generation runtime in the recently added last line of slurm log entries (see #525). The code no longer requires a separate text file with job IDs and runtimes from slurm emails.


mcoughlin (approved):
LGTM

PR #527, opened 2024-01-09
Exclude invalid files from inference dataset

bfhealy:
This PR excludes invalid files in feature generation output (e.g. `.running` files) from inference, avoiding an `ArrowInvalid` error.


mcoughlin (approved):
LGTM

PR #528, opened 2024-01-10
Check running jobs before feature generation job submission

bfhealy:
This PR modifies `tools/generate_features_job_submission.py` to count running jobs using `squeue` before submitting additional feature generation jobs. This allows each 48-hour run of the submission script to be run back-to-back with the next run (e.g. as a cron job), since the new instance of the script will have access to the current active job count.

The new code also makes more efficient use of real time by recognizing any failed jobs and queuing a new jobs to take their place. Previously any failures would restrict the max number of instances available for new jobs queued from the same script run. The user will still need to run another iteration of the script to re-run any failed jobs.

### DIFF
--- a/tools/generate_features_job_submission.py
+++ b/tools/generate_features_job_submission.py
@@ -267,6 +267,7 @@ if __name__ == '__main__':
     print('%d jobs remaining to queue...' % nchoice)
 
     if args.doSubmit:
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #538, opened 2024-02-15
Update config defaults with changes from paper revision

bfhealy:
This PR updates the config defaults to reflect Scope III paper revisions (specifically adding Gaia EDR3 parallax error as an ontological feature with median imputation).


mcoughlin (approved):
LGTM

PR #539, opened 2024-02-16
Update analysis plots notebook with revised stats/figures

bfhealy:
This PR updates the SCoPe data analysis plots notebook with the stats and plots used in the revised paper.


mcoughlin (approved):
LGTM

PR #542, opened 2024-02-21
Update workflows in advance of PyPI refactoring

bfhealy:
Opening this PR to update GH Actions workflows to test PyPI refactoring (#514). These tests will fail here, but once merged they should help evaluate the changes in #514. Currently, the workflows running on that PR are not the updated versions.


mcoughlin (approved):
LGTM

PR #545, opened 2024-02-23
Add PyPI publishing workflow

bfhealy:
This PR adds a workflow to publish scope releases to PyPI using poetry. The workflow runs when a commit is tagged with `v*.*.*`. See https://github.com/marketplace/actions/publish-python-poetry-package for more info.

Note: tests will continue to fail until #514 is merged after some additional testing/debugging.


mcoughlin (approved):
LGTM

PR #553, opened 2024-03-12
Fix path bug in inference job submission

bfhealy:
This PR fixes a bug in inference job submission by changing the slurm script path used by the code.


mcoughlin (approved):
LGTM

PR #560, opened 2024-03-28
Write json file containing list of available fields

bfhealy:
This PR modifies `combine_preds.py` to write a json file containing a list of all available fields. This list will be added to the Zenodo record to describe which fields can be downloaded.


mcoughlin (approved):
LGTM

PR #566, opened 2024-04-02
Allow user to specify fields to exclude from feat gen/inference

bfhealy:
This PR allows a user to list fields to exclude from feature generation/inference in the config file (under fields_to_exclude). This prevents duplication of effort even if completed field files are not present locally.

### DIFF
--- a/config.defaults.yaml
+++ b/config.defaults.yaml
@@ -2244,6 +2244,6 @@ inference:
     # - 853
   path_to_preds:
 
-# Some fields may already be comleted and should be excluded from subsequent feature generation/inference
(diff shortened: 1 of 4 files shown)


config.defaults.yaml
```diff
@@ -2243,3 +2243,7 @@ inference:
     # - 852
     # - 853
   path_to_preds:
+
+# Some fields may already be comleted and should be excluded from subsequent feature generation/inference
```
mcoughlin, line 2247: completed

mcoughlin (approved):
LGTM

PR #567, opened 2024-04-03
Add util to save LC count by field given catalog_completeness file

bfhealy:
This PR adds a util to save a file containing field-by-field light curve counts given a catalog_completeness file (see `tools.generate_features_slurm.check_quads_for_sources`) as input.


mcoughlin (approved):
LGTM

PR #570, opened 2024-04-04
Add code to generate combine-preds slurm script

bfhealy:
This PR adds code to generate a slurm script for `combine-preds` - the last step of the feature generation/inference workflow.


mcoughlin (approved):
LGTM

PR #571, opened 2024-04-05
Save accurate list of available field predictions

bfhealy:
This PR modifies `combine_preds.py` to save an accurate list of available preds.


mcoughlin (approved):
LGTM

PR #572, opened 2024-04-05
Add cython requirement for periodfind

bfhealy:
This PR adds a cython requirement to make `periodfind` installations easier on cluster resources.


mcoughlin (approved):
LGTM

PR #574, opened 2024-04-05
Update slurm script generators with NSCA Delta specifics

bfhealy:
This PR updates all slurm script generators to include specific lines for NSCA Delta. Note that while scope feature generation runs on Delta GPUs, at the moment it appears to take ~2x as long to complete compared to Expanse. This seems too slow for A100s but also too fast to be avoiding the GPUs altogether.


mcoughlin (approved):
LGTM

PR #575, opened 2024-04-08
Add check-quads-for-sources script from generate_features_slurm

bfhealy:
This PR installs `check-quads-for-sources` as a new script that runs its namesake function within `generate_features_slurm.py`


mcoughlin (approved):
LGTM

PR #577, opened 2024-04-09
Fix dtype bug in combine-preds field lists

bfhealy:
This PR fixes a bug that mixed int and str types in the fields lists of `combine_preds.py`. The saved list now contains integer fields numbers instead of strings, but the code converts back to `"field_N"` format when checking existing files saved in that format.


mcoughlin (approved):
LGTM

PR #578, opened 2024-04-10
Add example HPC directories, config, scripts and documentation

bfhealy:
This PR adds a directory called `hpc_files` that contains example scripts and files useful for running scope at scale on cluster resources. A new docs page is also added to explain these scripts, their use cases, and common troubleshooting issues. An example config file (sans tokens) is also included.

There are additionally some minor bug fixes and documentation updates associated with these scripts.


mcoughlin (approved):
Hopefully Daniel understands some of these :)

bfhealy:
So far so good - Daniel was able to produce a new field of classifications on his Expanse account!

PR #586, opened 2026-07-23
Agn features - DRW, EVT anomaly scores and PDRS 

zerozole:
# AGN / anomaly features for the SCoPe pipeline (`agn-features` branch)

## Motivation

SCoPe's feature engineering is tuned toward periodic and eclipsing variables.
This branch extends the feature set toward AGN and stochastic variability:
damped-random-walk (DRW) parameters, extreme-value-theory (EVT) anomaly
scores, and flare / high-activity segmentation. Light curves are thereby
marked for anomalies without touching any existing behavior: all new
features default to `include: false` (annotation-only), both new
feature-generation steps are behind opt-in flags, and no existing model,
feature file, or training set is affected.

Methods are adapted from two existing research codes:

- **PDRS** — (https://github.com/zerozole/Peak_Driven_Region_Segmentation) - peak-detection region segmentation of flares 
  (binned relative flux, gradient-guided region growing, saddle-point merging, region-median
  gate).
- **FLARE**  — (https://github.com/zerozole/FLARE) - DRW/OU baseline fitting
  with celerite and a peaks-over-threshold generalized Pareto calibration of
  the maximum standardized residual.

## What was added

| File | Change |
|---|---|
| `tools/featureGeneration/pdrs.py` | new: flare segmentation → `n_flares`, `flare_significance`, `flare_duty_cycle` |
| `tools/featureGeneration/drw.py` | new: celerite DRW/OU fit (RealTerm + JitterTerm, L-BFGS-B) → `drw_tau`, `drw_sigma`, `drw_max_z`, `drw_fit_success` |
| `tools/evt_calibration.py` | new: population-level GPD calibration → `anomaly_score`, `anomaly_flag` columns + frozen JSON calibration |
| `tools/generate_features.py` | `--doFlareDetection` / `--doDRW` flags (default off), computation hooked in after basic stats, hyperparameters read from config |
| `tools/generate_features_slurm.py` | forwards the two new flags |
| `config.defaults.yaml` | feature registration (`include: false`) + `feature_generation.pdrs/.drw/.evt` hyperparameter blocks |
| `requirements.txt`, `pyproject.toml` | `celerite>=0.4.2` dependency; `evt-calibration` console script |
| `tests/test_agn_features.py` | 22 synthetic-data tests (no Kowalski/GPU needed) |

`periodfind` is untouched.

## Method summary

**DRW features** (`--doDRW`): each light curve is fit with a celerite GP —
[...]

PR #587, opened 2026-08-03
Add per-band DRW and PDRS flare features to Rubin feature generation

zerozole:
# Add per-band DRW, EVT scores and PDRS flare features to Rubin feature generation

Follow-up to #586, which added the AGN/anomaly feature modules and wired them into the ZTF
entry point. This ports them into the Rubin entry point behind the same two flags,
`--doFlareDetection` and `--doDRW`. Covers DP1 and DP2, since `--release dp2` runs through
`generate_features_rubin.py` as well.

One file, 64 insertions, seven wiring sites: the import, `pdrs_params`/`drw_params` at the
config header, the two flags in the signature, the docstring, a new section 4c after 4b,
argparse, and arg passing in `main()`.

## Per band, not on the merged curve

The ZTF path never faced this choice. Rubin's `tme_dict[oid]['tme']` merges up to six bands
with the band index held separately, so the features could be computed either way.

They are computed per band because:

- Rubin's own `DiaObject` schema is per band throughout —
  `{u,g,r,i,z,y}_psfFlux{Ndata,Mean,Sigma,MAD,StetsonJ,Skew,Chi,...}`.
- Section 4b of this file already computes per-band statistics (`n_{band}`, `wstd_{band}`)
  explicitly "to avoid inflation from chromatic offsets."
- A DRW fit to the merged curve would read chromatic offsets and per-band amplitude
  differences as intrinsic red noise, biasing `drw_sigma` high.

## Output

42 columns per source — 6 bands × [3 PDRS + 4 DRW] — named `drw_tau_g`, `n_flares_r`, and so
on, following the existing `wstd_g` convention.

Both modules gate internally on epoch count and return fixed-key NaN dicts for curves that are
too short, so the column set is identical across sources regardless of band coverage. No manual
gating is needed, and the resulting feature table has a stable schema.

## No config changes

The hyperparameters registered in #586 under `feature_generation.pdrs` / `.drw` are
survey-agnostic, so nothing needed to be added to `config.defaults.yaml`. The epoch floor stays
tunable via `feature_generation.drw.min_epochs`.

## Testing

- Full suite: 236 passed, 12 skipped (the skips need `RUBIN_TAP_TOKEN`, DP1/EDP2 data on
  `/fred`, or the IRSA opt-in).
- Per-band behaviour verified against a synthetic multi-band light curve — correct column
  names, stable schema across differing band coverage, and NaN fill on short bands.

### DIFF
--- a/tools/generate_features_rubin.py
+++ b/tools/generate_features_rubin.py
@@ -32,7 +32,7 @@ from scope.utils import (
 )
 from scope.surveys.rubin import make_rubin_client
 from tools.get_rubin_ids import get_rubin_objects_by_cone, get_rubin_objects_from_file
-from tools.featureGeneration import periodsearch
+from tools.featureGeneration import periodsearch, pdrs, drw
 
 # Flush stdout/stderr after every print so SLURM logs update in real time
 sys.stdout.reconfigure(line_buffering=True)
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #588, opened 2026-08-08
Implement prepare-rubin-chunks, generate-features-rubin-slurm, combin…

james-freeburn:
Author: Claude
…e-rubin-features

These were documented in the Rubin DP1/DP2 user guides and README but had no entry points or scripts backing them. Adds:

- get_eligible_objects() on RubinLocalClient, RubinLocalDiaClient, and RubinLocalDP2Client to filter by minimum valid detections.
- tools/prepare_rubin_chunks.py: scan a local DP1/DP2 export and split eligible objects into chunk CSVs.
- tools/generate_features_rubin_slurm.py: generate a SLURM array script that runs generate-features-rubin per chunk, forwarding --release/ --use-dia/--flux-column so it works for both DP1 and DP2.
- tools/combine_rubin_features.py: merge per-chunk parquet outputs.
- Fix tools/get_rubin_ids.py importing the nonexistent scope.rubin module instead of scope.surveys.rubin, which broke get-rubin-ids and generate-features-rubin.
- Register the three new entry points in pyproject.toml and update the CLI reference table.


mcoughlin:
@james-freeburn can you fix the linting? You can use pre-commit install and then run it on your files.

mcoughlin (approved):
LGTM

